PR #1, opened 2022-10-11
Working Version

itehle:
This version should work on any machine that has the two required ROOT files, hadd.root and test_triggergeom.root.


bfonta:
Would you please squash your commits into one a couple for readability? You have many with the same message plus merge commits which should not be propagated to the upstream branch.

README.org
```diff
@@ -15,11 +15,15 @@ git clone git@github.com:bfonta/bye_splits.git
 
 I recommend creating a separate =conda= environment for running the code.
 #+BEGIN_SRC shell
-create -n <EnvName> python=3.9 pandas
+create -n <EnvName> python=3.9 pandas uproot pytables h5py
 conda install -c bokeh bokeh
 #+END_SRC
 
-Run ~conda activate <EnvName>~ to enter the environment and ~conda deactivate~ to exit it.
+Then run ~conda activate <EnvName>~ to enter the environment (~conda deactivate~ to exit at any time). In case of package conflicts, you can install each package using ~conda install -c conda-forge <PackageName>~. This ensures that every package will be installed from the conda-forge package channel, and will ensure there aren't inconsisties coming from cross-channel installations.
```
bfonta, line 22: typo in "inconsistency"

README.org
```diff
@@ -15,11 +15,15 @@ git clone git@github.com:bfonta/bye_splits.git
...
+
+You will also need to create two new directories, ~/data~ and ~/out~, on the same level as the second ~/bye_splits~ directory (parallel to ~/.git~ if cloned from this repository).
```
bfonta, line 24: Can you actually include those commands in the shell file? Something like ```mkdir -p ...```, using the already defined shell variables.

README.org
```diff
@@ -15,11 +15,15 @@ git clone git@github.com:bfonta/bye_splits.git
...
+
+You will need to place the =test_triggergeom.root= file in ~/data~ before running the reconstruction chain.
```
bfonta, line 26: You mean ```~data/~```, as this is not an absolute path (here I am being pedantic, I know ;) )

bye_splits/production/prod_params.py
```diff
@@ -26,7 +26,8 @@
     if htcondor:
         files_photons = glob.glob('/home/llr/cms/sauvan/DATA_UPG/HGCAL/Ntuples/study_autoencoder/3_22_1/SinglePhoton_PT2to200/GammaGun_Pt2_200_PU0_HLTWinter20_std_ae_xyseed/210430_091126/ntuple*.root')
     else:
-        files_photons = [ str(Path('/data_CMS') / 'cms' / os.environ['USER'] / 'TriggerCells' / 'hadd.root') ]
+        #files_photons = [ str(Path('/data_CMS') / 'cms' / os.environ['USER'] / 'TriggerCells' / 'hadd.root') ]
+        files_photons = [ str(Path('/data_CMS') / 'cms' / 'alves' / 'TriggerCells' / 'hadd.root') ]
```
bfonta, line 30: This is supposed to work even after my personal file is gone. I think I prefer the original version. As an alternative, you could throw an error message if the directory is not found; in this latter case you could use any default directory you like.

bfonta (changes requested):

bye_splits/production/prod_params.py
itehle: For that to work, the file would have to be USER/TriggerCells, but that isn't the case for an arbitrary user (for example, I don't even have a directory in /data_CMS/cms/). That's likely a problem on its own, but regardless the file should either 1) be kept in a shared location on the llr_servers so that anyone can access it or 2) created within the pipeline. Option (1) seems significantly easier for now, but you're right that it should probably be kept somewhere other than your personal folder.

bfonta:
Looking at [the commit](https://github.com/bfonta/bye_splits/compare/0d39214beaca27a42068e09c781547d7ea682caf..d42fa123b54da03bc5c2b7fcb284468a8ac6e7c7) you introduced, we can see that:

1) You did not base it on the latest version of my ```main``` branch and many conflicts are due to this mismatch (some updates are being undone)
2) [The modification](https://github.com/bfonta/bye_splits/commit/d42fa123b54da03bc5c2b7fcb284468a8ac6e7c7) you added to the ```comm``` variable surely does not run (there are semicolons missing)

This is what I suggest: get the ```HEAD``` of my ```main``` branch and rebase your commit on top of it (solving all the conflicts that will arise). Once your commit is conflict-free, and once you tested all the changes you introduced, open a new PR with that single commit, that I should be able to merge directly. Please do not open a PR with untested changes.

PR #2, opened 2022-10-12
Master

itehle:
Squashed minor commits.


itehle:
Realizing that the commits were _not_ squashed, and instead I may have made it worse. I attempted to squash via git rebase -i HEAD~10 and then letting pick-->squash for the README updates that happened after "Compatibility Tests." Apparently this didn't work, but I'm not sure why...

PR #3, opened 2022-10-18
Preliminaries

itehle:
Should be okay now.


bye_splits/tasks/EnPerDelR.py
```diff
@@ -0,0 +1,76 @@
+import os
```
bfonta, line 1: This file corresponds to an independent study, and not to a reconstruction task. Could you please move it to a folder called ```scripts/``` (for instance) so that the repository continues well organized?

bye_splits/tasks/EnPerDelR.py
```diff
@@ -0,0 +1,76 @@
...
+import numpy as np
+import pandas as pd
+import h5py
+
+import math
+
+def energy(pars, **kw):
+
+    coefs = np.linspace(0.00,0.05)
+    energies = np.array([0])
```
bfonta, line 22: The variable ```energies``` is not being used.

bye_splits/tasks/EnPerDelR.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # This will create a .hdf5 file containing the cluster energy information corresponding to each coefficient
+    # By default this file is assumed to be there
+    if kw['ReInit']:
+        for coef in coefs:
+            clust_params = params.cluster_kw
```
bfonta, line 28: There is no need to define this variable per iteration; better do it outside.

bye_splits/utils/parsing.py
```diff
@@ -29,3 +29,4 @@ def add_parameters(parser, meta=False):
                         choices=('max_energy', 'min_distance'),
                         default='min_distance', type=str,
                         help='Clustering algorithm applied.')
+    #parser.add_argument('--coef', help='Value of the CoefficentA, or DeltaR=(x,y)/z. Default is 0.015', default=0.015, type=float)
```
bfonta, line 32: If you don not use it, could you remove this comment?

bfonta (changes requested):
Check the comments below. In addition, hadn't you introduced some changes to the ```README``` file?

bye_splits/tasks/EnPerDelR.py
itehle: Yeah no worries, I'll change the folder and name. As for the comments, that's just because the file is still being written and I'm trying things out, but apologies, I shouldn't have pushed before I have a working version (well, this version runs, but it doesn't really do anything).
As for the white spaces, yeah, I need to figure out why it's creating them. I'll try to fix all of the above before the next commit.

bye_splits/tasks/EnPerDelR.py
itehle: It will be (it's used and working in the commented out section).

bye_splits/tasks/EnPerDelR.py
itehle: Good catch, thanks!

bye_splits/utils/parsing.py
itehle: I need to confirm that I really won't need it, because I wrote that for one plan for the energy optimization, but I think I found a better way that won't require it, hence commenting it out. But I don't want to remove it entirely until I know for sure that this new way will do the trick. Again, apologies for the PR before the new script is completed, still don't know what the best git etiquette is. Working on it.

itehle:
The changes to the README were unnecessary, your version was/is fine, so I decided to drop them when I did the new fork.

PR #4, opened 2022-10-25
setting up crab config directory + including link to link mapping tools

portalesHEP:
@bfonta FYI 

The "link mapping tools" are require to define a mapping adapted to the BC+STC algorithm using the "old" V2 geometry. BC+STC can't yet be used with the current V3 geometry; the adaptation is apparently WIP.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "bye_splits/crab/realistic-signaldriven-elinks"]
+	path = bye_splits/crab/realistic-signaldriven-elinks
+	url = git@github.com:portalesHEP/realistic-signaldriven-elinks.git


bfonta:
Ok, then let us go for ```subtree``` as long as you can use it, as it also brings a hard copy of the files, and let us use your fork. In my case the repository is sitting on ```CMSSW_12_5_0_pre1``` and ```git --version``` returns ```2.35.3```. Subtrees are available from ```1.7.11``` onwards.
Can you please update the PR? [This file](https://gist.github.com/SKempin/b7857a6ff6bddb05717cc17a44091202) has some nice quick introduction to subtrees.

bfonta:
I don't remember why we discussed this, but we clearly never used it. Closing it for now. 
@portalesHEP Feel free to comment.

bfonta:
After a quick chat with @isehle, we realized this PR is most likely outdated. We will create a PR with an updated crab file soon. It would be nice to include some instructions on how to run it, including CMSSW version and available options.

PR #7, opened 2023-01-19
update electron event numbers

mchiusi:
update electron event numbers


bfonta (approved):
Brilliant piece of code!

PR #10, opened 2023-01-25
[WIP] Include module sums

bfonta:
Main updates:
- [x] Include modules sums skimming
- [x] Add simple progress information to skimming
- [ ] Include color mapping for module sums
- [ ] Deal with columns having a different number of elements per event

Unrelated updates:
- [x] Add ```genpart_pos*``` variables (x, y and z: gen particle position per layer) as a vector of vectors of floats: one vector (currently of length 50) per generated particle
- [x] Throw error if an event (default or chosen by the user) does not exist in the dataset


bfonta:
@isehle @mchiusi Is there any update on module sums?
This PR is so outdated that we might simply create a new one when time comes.

mchiusi:
> @isehle @mchiusi Is there any update on module sums?
> This PR is so outdated that we might simply create a new one when time comes.

Yes, that's a good idea. Thank you for cleaning up the list of outdated PRs.

PR #11, opened 2023-01-25
Add event display (2d & 3d) dashApp

mchiusi:
The added files allow to visualise events from `/eos/user/b/bfontana/FPGAs/new_algos/skim_small_electrons_0PU_bc_stc_hadd.root` `ROOT` file.
Similar files can be used: the needed variables are efficiently stored in `bye_splits/data_handle/config.yaml`.  It's possible to change the input file by changing the `EOSDataFolder` variable in `bye_splits/utils/params.py`.

The main script is `bye_splits/plot/display_plotly/main.py`. Its Python execution allows to launch a dashApp running on `http://llruicms01.in2p3.fr:8004/`.

Files to get data from the `ROOT` file are stored in `bye_splits/data_handle` folder.
Files to analyse each single event are in `bye_splits/plot/display_plotly/` folder. The definition of the figures and production of 2 and 3D plots is present as well.

From the web app it's possible to:
- visualise 3D events;
- visualise 2D events, layer by layer;

Then, for each 2D or 3D case:
- to change event (if not specified, an event is chosen from the event list present in the `ROOT` file). If the chosen event is not found an error will occur, but the GUI is still on;
- to modify the energy threshold (in MIP_T);
- to visualise the clustered and non clustered trigger cells of each event.

This is how an event should appear in the 3D visualisation mode:
Screenshot 2023-01-25 at 18 36 07


bfonta:
Very nice addition, we are getting quite close of having a workable event display! Some comments: 

- Have you checked the code also works for photons samples? Also:
    1. it is better not to include a path to a data input file that will be likely changed in the future
    2. the file is not part of the repository, so it might be better not to mention it explicitly
    3. the path itself will likely change in the future
- In the same line, you mention many files by name and how they are used. As soon as those files are modified, this PR's text will become outdated. If you really want to include mentions to specific code elements use a link _to the last commit you are including_ (not to a branch, as this also gets updated often)
- Could you please add units to the axis labels? Where can we find the axis (plus label) for the color code?
- Can you include a picture showing the energy threshold slider you mention?
- I would like to see the above for the full geometry, as a check that everything is as expected (like what was done in #5 )

bye_splits/data_handle/event.py
```diff
@@ -10,6 +10,7 @@
 
 import yaml
 import uproot as up
+import numpy as np
```
bfonta, line 13: ```numpy``` here is not required, as the method also exists in ```awkward```

bye_splits/data_handle/event.py
```diff
@@ -50,6 +52,14 @@ def cache_events(self, events):
             self.cache = pd.concat([self.cache, ds], axis=0)
         #self.cache = self.cache.persist() only for dask dataframes
 
+    def get_event_number(self):
```
bfonta, line 55: What information do you want exactly? If you want all event id's present in the input files, then the function should be called ```get_event_numbers```, and please include a docstring. I also suggest you to look into python properties (used for getters and setters).

bye_splits/data_handle/event.py
```diff
@@ -65,6 +75,9 @@ def provide_event(self, event):
         ret = ret.apply(pd.Series.explode).reset_index(drop=True)
         return ret
     
+    def provide_event_number(self):
```
bfonta, line 78: You pick events at random? What is the point?

bye_splits/data_handle/geometry.py
```diff
@@ -272,8 +271,8 @@ def select(self):
         """Performs data selection for performance."""
         with up.open(self.inpath) as f:
             tree = f[ os.path.join('hgcaltriggergeomtester', 'TreeTriggerCells') ]
-            if self.logger is not None:
-                self.logger.debug(tree.show())
+            #if self.logger is not None:
```
bfonta, line 274: Why did you comment the above? The logger can be useful for debugging.

bye_splits/plot/display_plotly/event_processing.py
```diff
@@ -0,0 +1,143 @@
...
+                                            showbackground=True,
+                                            zerolinecolor="white",),
+                                 yaxis=dict(backgroundcolor="rgba(0,0,0,0)",
+                                            gridcolor="white",
+                                            showbackground=True,
+                                            zerolinecolor="white"),
+                                 zaxis=dict(backgroundcolor="rgba(0,0,0,0)",
+                                            gridcolor="white",
+                                            showbackground=True,
+                                            zerolinecolor="white",),),
```
bfonta, line 67: Since all the ```dict```s are the same you could define just one on top; if only part of the parameters are common you can use the ```**dict``` syntax.

bye_splits/plot/display_plotly/main.py
```diff
@@ -0,0 +1,94 @@
...
+
+from dash import Dash, dcc, html, Input, Output, State
+from dash.exceptions import PreventUpdate
+import argparse
+
+import numpy as np
+import pandas as pd
+import event_processing as processing
+
+parser = argparse.ArgumentParser()
```
bfonta, line 19: In general is good practice to put the parser in a ```if __name__ == "__main__":``` block.

bye_splits/plot/display_plotly/main.py
```diff
@@ -0,0 +1,94 @@
...
+            df_no_cluster = df_layer.loc[df_layer.tc_cluster_id == 0]
+            df_cluster    = df_layer.loc[df_layer.tc_cluster_id != 0]
+            fig = processing.set_2dfigure(df_cluster)
+            fig = processing.update_2dfigure(fig, df_no_cluster)
+        else:
+            fig = processing.set_2dfigure(df_layer)
+        return fig, u'Layer {} displayed'.format(slider_value) 
+    raise PreventUpdate
+
+if __name__ == '__main__':
```
bfonta, line 91: Very likely we will need to adapt the port and host names. Can you please add them in the parser, and use the values above as default (for the moment)?

bye_splits/plot/display_plotly/plot_event.py
```diff
@@ -0,0 +1,39 @@
...
+    listdata = []
+    for j,i in enumerate(array_data):
+        x1 = np.append(i[0][0][0],i[0][0][0][0])
+        y1 = np.append(i[1][0][0],i[1][0][0][0])
+        z1 = np.array(int(len(x1)+1) * [i[2]])
+        datum = go.Scatter3d(x=x1, y=y1, z=z1, opacity=opacity,mode="lines",
+                            surfaceaxis=surfaceaxis,surfacecolor=i[5],marker=dict(color="black"),
+                            text=('Energy: '+str(i[6]))
+                            )
+        listdata.append(datum)
```
bfonta, line 26: This looks horrendously inefficient. Are we sure there is no way to directly feed a pandas dataframe to a ```Scatter3D``` call? I would not mind reworking a bit the dataframe structure if it helps (the current structure is ```bokeh``` friendly, but we could add a ```plotly``` friendly version as well).

bye_splits/plot/display_plotly/plot_event.py
```diff
@@ -0,0 +1,39 @@
...
+
+def produce_2dplot(df, opacity=1):
+    array_data = df[['diamond_x', 'diamond_y', 'tc_x', 'tc_y', 'colors','mipPt']].to_numpy()
+    listdata = []
+    for j,i in enumerate(array_data):
+        x1 = np.append(i[0][0][0],i[0][0][0][0]) 
+        y1 = np.append(i[1][0][0],i[1][0][0][0])
+        datum = go.Scatter(x=x1, y=y1, opacity=opacity,mode="lines",fill='toself', fillcolor=i[4],
+                          line_color='black',marker_line_color="black",  text=('Energy: '+str(i[5])))
+        listdata.append(datum)
```
bfonta, line 37: Same as before.

bfonta (commented):
- Could you add docstrings to most functions, and some comments?
- Sometimes you reuse some parameters; try to define those in one place only.

bye_splits/data_handle/event.py
mchiusi: AttributeError: module 'awkward' has no attribute 'unique'

bye_splits/data_handle/event.py
mchiusi: Name changed and docstring added.

bye_splits/data_handle/event.py
mchiusi: As explained above when you click the button 'change event', a random event is chosen among the list returned by `get_event_numbers`.
In this way you can visualise some events and see for example if the clustering algorithm is working fine or not. In the next future JB and I plan to random select some of the 'pathological events'.

bye_splits/data_handle/geometry.py
mchiusi: As you prefer :)

bye_splits/plot/display_plotly/event_processing.py
mchiusi: Right. Correct

bye_splits/plot/display_plotly/plot_event.py
mchiusi: At the moment I don't see any other way to do it. You have to define each polygon and add its trace to the figure.
Any suggestion is of course well accepted!

bye_splits/data_handle/event.py
bfonta: You are definitely right; I got confused with [this PR](https://github.com/scikit-hep/awkward/issues/1104), which is still open.
However, for this case, you should be able to do what you want with [pandas.unique](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.unique.html).

bye_splits/data_handle/event.py
bfonta: Ok, thank you. Looking forward to see those in your next commit.

bye_splits/data_handle/event.py
bfonta: I see, it is an interesting addition I had not though of. Please then change the title of the buttom for something including the word "random" (at least for the moment).

bye_splits/plot/display_plotly/plot_event.py
bfonta: Can't you do something [along these lines](https://github.com/bfonta/bye_splits/blob/69d9fb94cde562ce694d80a8b78f06f071e6b0a2/bye_splits/plot/join/app.py#L36-L45)?
Perhaps @itehle has some suggestion?

bye_splits/data_handle/event.py
```diff
@@ -50,6 +52,14 @@ def cache_events(self, events):
             self.cache = pd.concat([self.cache, ds], axis=0)
         #self.cache = self.cache.persist() only for dask dataframes
 
+    def get_event_number(self):
+        ds = ak.from_parquet(self.outpath)
+        ds = ak.to_dataframe(ds)
```
bfonta, line 57: There is no need to convert to a pandas dataframe when you only want the event numbers, which you can access with ```ds.event```.

bye_splits/data_handle/event.py
mchiusi: Done

bye_splits/data_handle/event.py
mchiusi: Righ

bye_splits/data_handle/event.py
mchiusi: Solved

bye_splits/plot/display_plotly/main.py
mchiusi: Good idea, thanks!

bye_splits/plot/display_plotly/main.py
mchiusi: parser in `if __name__ == "__main__":` block now!

bye_splits/plot/display_plotly/plot_event.py
mchiusi: I'm using `plotly.graph_objects`. The only two options to do it through this library are reported [here](https://plotly.com/python/shapes/):
- adding more traces
- interrupting the series with `None`

Maybe we could change library.. we can discuss

bfonta: Looking at the link you provided, indeed using ```None``` seems reasonable (it won't work with "holes" within the same shape, but we should not have those :crossed_fingers:). 
Given the above, I do not think we need to change the current data format, that works out of the box and would require a small transformation for ```plotly```. 

I would modify [this function](https://github.com/bfonta/bye_splits/blob/470158873f3924514269094fdf569100005dfaab/bye_splits/data_handle/geometry.py#L116), adding the ```library='plotly'``` possibility, and performing the necessary modifications to ```diamond_x/y``` and ```hex_x/y```. Namely, you likely do not want the list of lists per row (like [this](https://github.com/bfonta/bye_splits/blob/470158873f3924514269094fdf569100005dfaab/bye_splits/data_handle/geometry.py#L240) which, by the way, is quite slow and raising issues during deployment). Feel also free to add a conversion to a uni-dimensional list with interspersed ```None```s, making sure it is only done when ```library='plotly'```.

Let me know what you think. @itehle feel also free to comment, as this addition might be useful for you sooner or later.

mchiusi: @bfonta I changed a bit the geometry functions (see last commit).

bye_splits/data_handle/event.py
```diff
@@ -65,10 +72,14 @@ def provide_event(self, event):
...
         return ret
     
+    def provide_event_numbers(self):
+        return np.random.choice(self.event_numbers)
+
     def select(self):
         adir = 'FloatingpointMixedbcstcrealsig4DummyHistomaxxydr015GenmatchGenclustersntuple'
         atree = 'HGCalTriggerNtuple'
 
+        print(self.inpath)
```
bfonta, line 82: please remove

bye_splits/data_handle/geometry.py
```diff
@@ -289,7 +295,7 @@ def select(self):
             #data = data.drop_duplicates(subset=[self.var.cu, self.var.cv, self.var.l])
             data[self.var.wv] = data.waferv
             data[self.var.wvs] = -1 * data.waferv
-            data[self.var.c] = "#8a2be2"
+            data[self.var.c] = np.full(len(data), "#8a2be2")
```
bfonta, line 298: IIUC, ``` #8a2be2  ``` is enough, as the broadcasting is done automatically both by ```pandas``` and by ```awkward```, as shown below:

```shell
(base) bruno@pc:~  $ python
Python 3.9.5 (default, Jun  4 2021, 12:28:51) 
[GCC 7.5.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pandas as pd
>>> df = pd.DataFrame({'x': [0,1,2], 'y': [4,5,6]})
>>> df
   x  y
0  0  4
1  1  5
2  2  6
>>> df['z'] = 7.
>>> df
   x  y    z
0  0  4  7.0
1  1  5  7.0
2  2  6  7.0
>>> import awkward as ak
>>> df2 = ak.Array({'x': [0,1,2], 'y': [4,5,6]})
>>> df2
<Array [{x: 0, y: 4}, {...}, {x: 2, y: 6}] type='3 * {x: int64, y: int64}'>
>>> df2['z'] = 7.
>>> df2
<Array [{x: 0, y: 4, z: 7}, ..., {x: 2, ...}] type='3 * {x: int64, y: int64...'>
>>> df2
<Array [{x: 0, y: 4, z: 7}, ..., {x: 2, ...}] type='3 * {x: int64, y: int64...'>
>>> df2['z']
<Array [7, 7, 7] type='3 * float64'>
```

bye_splits/plot/display_plotly/main.py
```diff
@@ -0,0 +1,102 @@
...
+def make_graph(submit, data, layer, cluster, mip, slider_value, page):
+    df = pd.read_json(data, orient='records')
+    df_sel = df.loc[df.mipPt >= mip]
+    
+    if layer == 'layer selection':
+        df_sel = df_sel[(df_sel.layer >= slider_value[0]) & (df_sel.layer <= slider_value[1])]
+    
+    if cluster == 'cluster':
+        df_no_cluster = df_sel.loc[df_sel.tc_cluster_id == 0]
+        df_cluster    = df_sel.loc[df_sel.tc_cluster_id != 0]
```
bfonta, line 63: I believe ```loc``` is not required.

bye_splits/plot/display_plotly/main.py
```diff
@@ -0,0 +1,102 @@
...
+              [Input('dataframe', 'data'), Input('slider-range', 'value')],
+              [State('tc-cl', 'value'), State('mip', 'value'), State('page', 'value')])
+def make_graph(data, slider_value, cluster, mip, page):
+    df = pd.read_json(data, orient='records')
+    df_sel = df.loc[df.mipPt >= mip]
+
+    df_sel = df_sel[df_sel.layer == slider_value[1]]
+    if cluster == 'cluster':
+        df_no_cluster = df_sel.loc[df_sel.tc_cluster_id == 0]
+        df_cluster    = df_sel.loc[df_sel.tc_cluster_id != 0]
```
bfonta, line 86: same as before

bye_splits/plot/display_plotly/plot_event.py
```diff
@@ -0,0 +1,47 @@
+# coding: utf-8
+
+_all_ = [ 'EventData' ]
```
bfonta, line 3: remove ```'EventData'```, since it refers to a different file; please read about ```__all__``` in general, might be useful to know

bye_splits/utils/params.py
```diff
@@ -17,7 +17,7 @@
 MinPhi = -np.pi
 MaxPhi = +np.pi
 LocalDataFolder = 'data/new_algos'
-EOSDataFolder = '/eos/user/b/bfontana/FPGAs/new_algos/'
+EOSDataFolder = '/eos/user/m/mchiusi/visualization/'#'/eos/user/b/bfontana/FPGAs/new_algos/'
```
bfonta, line 20: Keep it as it was or, much better, find a general approach that works for everybody. Probably the first, as our time is limited, but you decide!

bfonta (changes requested):
Essentially everything is fine. Just add the remaining details and we can merge this.

bfonta (approved):
For the future: the full geometry in 3D is unbearably slow (much slower than for 2D, as expected, given the larger number of rendered objects and the naturally inefficient way of adding objects to a ```plotly``` graph). The 2D visualization is also a bit too slow to be used comfortably when all trigger cells are on.
It is to see whether the event displays remain usable with pile-up; this will likely be extremely dependent on the initial trigger cell energy cut. The visualization for events without pile-up is fine, so we will for the moment avoid early optimizations that are hopefully not required.

PR #17, opened 2023-02-03
DashApp for Cluster Size Studies

itehle:
Resolved merge conflicts before opening the PR so this should be easy


bfonta:
Please do the following before we start the review:
+ Remove all spurious files: dot files, vscode-related, ...
+ Squash all or most commits, providing a clear and concise message of what each achieves
+ Add a good description of what this PR is trying to achieve
+ You are modifying almost all files of the repository: this is due, at least in part, to the way your text editor processes blank spaces and tabs; I cannot accept modifications including only those, so please change the settings of your text editor

itehle:
For some reason I do not have the option to directly reply to [your comment](https://github.com/bfonta/bye_splits/pull/17#issuecomment-1416182455) so I will do so here:

1. My latest commit has removed all dot files, as well as other unnecessary files.
2. I have squashed my latest commits, but I am currently receiving an error "Unable to squash. Squashing replays all commits up to the last one required for the squash. A merge commit cannot exist among those commits." when I attempt to squash others. Likely the simplest solution would be to let the PR squash all commits into one, though I am not sure this can be done now or if I need to close this PR and make a new one. In any case this would also have the advantage of not squashing the commits on my own fork, because I actually find the majority of them useful benchmarks for myself (though I agree a few of them do need to be eventually squashed).
3. Yes, apologies, I meant to write a different description than I did.
4. VSCode doesn't let you ignore whitespace, which I know is very annoying. However, you can ignore it on GitHub in the PR, correct?

PR #20, opened 2023-02-17
Share dash

isehle:
### PR Description

Integrates the Dash App used for cluster studies, primarily showing the effect of maximum cluster radius size (henceforth referred to as "coefficients") on reconstruction efficiency. The primary script,```bye_splits/scripts/cluster_size/run_cluster_size.py```, runs the Reconstruction Chain using the parameters set in ```bye_splits/scripts/cluster_size/cl_size_params.yaml```. The clustering step is repeated for a range of coefficients whose values are also set in the configuration file—details below. You can (and probably should) use the script ```run_cluster_size.sh``` which acts as a wrapper for the Python script, passing some parameters that we consider defaults that are nonetheless not the default options used in the main reconstruction chain.

To run the cluster size study, use ```bash run_cluster_size.sh <user>``` where ```<user>``` corresponds to your lxplus username. The output files currently land in a directory ```<base_dir>/data/new_algos/```, though I am currently working on implementing an option to send the files directly to your ```eos/``` directory, assuming the structure:

```
├── /eos/user/<first_letter>/<username>
│                                   ├── out
│                                   ├── data
│                                   │   ├──PU0
│                                   │   │   ├──electrons
│                                   │   │   ├──photons
│                                   │   │   ├──pions
│                                   │   ├──PU200
│                                   │   │   ├──electrons
│                                   │   │   ├──photons
│                                   │   │   ├──pions
```

Note that the current implementation for reading in the data, which relies on Pandas DataFrames, is far, far too slow to run this study using pile up samples. For reference, having verified on a subset of events from the two cases, the average number of active trigger cells per event without pile up is of order 10^2, while with pile up it is of order 10^4. This difference causes an immense lag when reading in the data, specifically at [this line](https://github.com/bfonta/bye_splits/blob/e509e31d9252d1a372ca9e054d61af250c35c88a/bye_splits/data_handle/event.py#L156). For verification purposes however, you can add an optional ```entry_stop``` parameter to this function call to see a subset of entries; this tends to work fine if ```entry_stop~100```.

To run the dash up, use the same syntax as before ```bash run_cluster_app.py <user>```, changing the parameters in ```bye_splits/plot/display_clusters/config.yaml``` if needed. The app should not require you to have run the cluster size study yourself if you have read access to my directory ```/eos/user/i/iehle/data/```; it checks for the input files in your directory first before checking mine. All outputs go to your directory.

Additions to ```parsing.py```:

1. ```--user``` : lxplus username
2. ```--cluster_studies``` : boolean, whether or not to run ```cluster_size.py```. Default is false.

Addition to ```common.py```:
The ```fill_path()``` module was hard-written to assume your local directory. I updated this to use the local directory (defined in ```params.py```) as default (so this shouldn't break anything), but the user now has the option to pass in the optional parameter ```data_dir``` if they want to specify a different path.

### How has this been tested?

Running either of the main scripts (```run_cluster_size.sh``` and ```run_cluster_app.sh```) works as described above without producing errors, provided the input information is set correctly. They are built to run quite independently from the rest of the script, so there should be not any compatibility issues. It is however important that you log in to ```/eos``` before attempting to run the scripts as they require read access to the files hosted there. Each page of the DashApp create ```.hdf5``` files for parameter combinations that haven't yet been tested, and this creation can be rather slow (order ~5 minutes to update the page). However once the files have been created, it simply has to read them, so the speed at which the app is updated will greatly improve over time.

### Attachments

N/A

### Breaking changes

1. The lambda function for the EOSStorage variable, defined [here](https://github.com/bfonta/bye_splits/blob/e509e31d9252d1a372ca9e054d61af250c35c88a/bye_splits/utils/params.py#L21), has been [updated](https://github.com/isehle/bye_splits/blob/d3fca8462dc833fea9ea71d04c98fc04386fc83a/bye_splits/utils/params.py#L24) to accept the trailing directory as a second argument.

2. In the data handling scripts, I have added a ```prod_key``` variable to the BaseData class initialization. The current implementation is hardcoded to read ```bye_splits/production/prod_params.yaml``` from ```CfgPaths['prod']``` in ```bye_splits/utils/params.py```. This change allows you to specify which key you pass and thus which configuration file is loaded. The changes involve adding the ```prod_key``` variable to the lines [here](https://github.com/bfonta/bye_splits/blob/e509e31d9252d1a372ca9e054d61af250c35c88a/bye_splits/data_handle/base.py#L55), [here](https://github.com/bfonta/bye_splits/blob/e509e31d9252d1a372ca9e054d61af250c35c88a/bye_splits/data_handle/data_handle.py#L19), [here](https://github.com/bfonta/bye_splits/blob/e509e31d9252d1a372ca9e054d61af250c35c88a/bye_splits/data_handle/data_handle.py#L52), [here](https://github.com/bfonta/bye_splits/blob/e509e31d9252d1a372ca9e054d61af250c35c88a/bye_splits/data_handle/data_handle.py#L64), [here](https://github.com/bfonta/bye_splits/blob/e509e31d9252d1a372ca9e054d61af250c35c88a/bye_splits/data_handle/event.py#L22) and [here](https://github.com/bfonta/bye_splits/blob/e509e31d9252d1a372ca9e054d61af250c35c88a/bye_splits/data_handle/event.py#L24).

3. I have added a key ```particles``` under the ```selection``` criteria of ```bye_splits/production/prod_params.yaml```. The data handling scripts were previously hardcoded to read only information from ```photons``` unless the user manually changed those scripts. In particular, [this line](https://github.com/bfonta/bye_splits/blob/e509e31d9252d1a372ca9e054d61af250c35c88a/bye_splits/data_handle/data_handle.py#L20) in ```data_handle.py``` will no longer take the "particles" variable as it will be included in the dictionary loaded in the previous line. This of course requires a minor change of the EventDataParticle factory; you can see how I did so [here](https://github.com/isehle/bye_splits/blob/64f92913363f38534e05872f32f2ad679e74047f/bye_splits/data_handle/data_handle.py#L77).

### Checklist


bye_splits/scripts/pages/eff_page.py
```diff
@@ -0,0 +1,291 @@
...
+    en_list = [0]
+    en_bin_size = perc*(df[norm].max() - df[norm].min())
+    if not perc==1.:
+        current_en = 0
+        for i in range(100):
+            match_column = df.loc[df[norm].between(current_en, (i+1)*en_bin_size, 'left'), 'matches']
+            if not match_column.empty:
+                try:
+                    eff = float(match_column.value_counts(normalize=True))
+                except:
```
bfonta, line 63: same as before

bye_splits/scripts/pages/eff_page.py
```diff
@@ -0,0 +1,291 @@
...
+                    eff = match_column.value_counts(normalize=True)[True]
+                eff_list.append(eff)
+                current_en += en_bin_size
+                en_list.append(current_en)
+    else:
+        match_column = df.loc[df[norm].between(0, en_bin_size, 'left'), 'matches']
+        if not match_column.empty:
+            try:
+                eff = float(match_column.value_counts(normalize=True))
+            except:
```
bfonta, line 73: same as before

bye_splits/scripts/pages/eff_page.py
```diff
@@ -0,0 +1,291 @@
...
+        coef_str = 'coef_{}'.format(str(coef).replace('.','p'))
+    else:
+        coef_str = coef
+    if coef_str not in file.keys():
+        coef_list = [float(re.split('coef_',key)[1].replace('p','.')) for key in file.keys()]
+        new_coef = closest(coef_list, coef)
+        coef_str = '/coef_{}'.format(str(new_coef).replace('.','p'))
+    return coef_str
+
+# Goal: Write function that finds and returns the DataFrames corresponding to a chosen coefficient in these files without explicitly referencing them
```
bfonta, line 89: Instead of comments, add this as a [docstring](https://peps.python.org/pep-0257/).

bye_splits/scripts/pages/eff_page.py
```diff
@@ -0,0 +1,291 @@
...
+def get_dfs(init_files, coef):
+    df_dict = dict.fromkeys(init_files.keys(),[0.0])
+    for key in init_files.keys():
+        if len(init_files[key])==1:
+
+            File = pd.HDFStore(init_files[key][0],'r')
```
bfonta, line 95: style detail (ideally to be enforced, but there is no time for everything): variables start with a lower-case letter

bye_splits/scripts/pages/eff_page.py
```diff
@@ -0,0 +1,291 @@
...
+    #file_name = start+re.split('gen_cl3d_tc',init_files['photons'][0])[1]
+    #file_path = common.fill_path(file_name, **pars)
+
+    file_path = init_files['photons'][0]
+    
+    # testing
+    print("\nFilepath: {}\n".format(file_path))
+
+    with pd.HDFStore(file_path, 'r') as File:
+        keys = File.keys()
```
bfonta, line 125: please remove the commented part

bye_splits/scripts/pages/eff_page.py
```diff
@@ -0,0 +1,291 @@
...
+
+def display_color(coef, eta_range, normby):
+    df_by_particle = get_dfs(input_files, coef)
+    phot_df = df_by_particle['photons']
+    pion_df = df_by_particle['pions']
+
+    phot_df = phot_df[ (phot_df['genpart_exeta'] > eta_range[0]) ]
+    pion_df = pion_df[ (pion_df['genpart_exeta'] > eta_range[0]) ]
+    phot_df = phot_df[ (phot_df['genpart_exeta'] < eta_range[1]) ]
+    pion_df = pion_df[ (pion_df['genpart_exeta'] < eta_range[1]) ]
```
bfonta, line 188: You can apply two selections in one go with the ```(...) & (...)``` syntax. You can also use the ```.``` accessor instead of ```['']```.

bye_splits/scripts/pages/eff_page.py
```diff
@@ -0,0 +1,291 @@
...
+                        'pions': [0.0]}
+    for coef in coefs[1:]:
+        dfs_by_particle = get_dfs(input_files, coef)
+        phot_df = dfs_by_particle['photons']
+        pion_df = dfs_by_particle['pions']
+
+        phot_df = phot_df[ (phot_df['genpart_exeta'] > eta[0]) ]
+        pion_df = pion_df[ (pion_df['genpart_exeta'] > eta[0]) ]
+        phot_df = phot_df[ (phot_df['genpart_exeta'] < eta[1]) ]
+        pion_df = pion_df[ (pion_df['genpart_exeta'] < eta[1]) ]
```
bfonta, line 226: same as before

bye_splits/scripts/pages/eff_page.py
```diff
@@ -0,0 +1,291 @@
...
+
+    fig.update_yaxes(type="log")
+
+    fig.update_layout(title_text="Efficiency/{}".format(normby), yaxis_title_text=r'$Eff (\frac{N_{Cl}}{N_{Gen}})$')
+
+    return fig, dbc.Table.from_dataframe(glob_effs)
+
+def write_eff_file(norm, coefs, eta, file):    
+    effs_dict = {'photons': [0.0],
+                        'pions': [0.0]}
```
bfonta, line 217: misalignment

bye_splits/scripts/pages/eff_page.py
```diff
@@ -0,0 +1,291 @@
...
+        pion_df = pion_df[ (pion_df['genpart_exeta'] > eta[0]) ]
+        phot_df = phot_df[ (phot_df['genpart_exeta'] < eta[1]) ]
+        pion_df = pion_df[ (pion_df['genpart_exeta'] < eta[1]) ]
+
+        if norm=='Energy':
+            phot_eff, _ = binned_effs(phot_df, 'genpart_energy', 1.)
+            pion_eff, _ = binned_effs(pion_df, 'genpart_energy', 1.)
+        else:
+            phot_eff, _ = binned_effs(phot_df, 'genpart_pt', 1.)
+            pion_eff, _ = binned_effs(pion_df, 'genpart_pt', 1.)
```
bfonta, line 233: alternative: 

```python
binned_var = 'genpart_energy' if norm == 'Energy' else 'genpart_pt'
phot_eff, _ = binned_effs(phot_df, binned_var, 1.)
pion_eff, _ = binned_effs(pion_df, binned_var, 1.)
```

bye_splits/scripts/pages/eff_page.py
```diff
@@ -0,0 +1,291 @@
...
+
+    if not os.path.exists(filename):    
+        effs_by_coef = write_eff_file(normby, coefs, eta_range, filename)
+    
+    else:
+        try:
+            with pd.HDFStore(filename, "r") as glob_eff_file:
+                effs_by_coef = glob_eff_file['/Eff'].to_dict(orient='list')
+                print(f"\nEffs: {effs_by_coef}")
+        except:
```
bfonta, line 264: same as before

bye_splits/scripts/pages/eff_page.py
```diff
@@ -0,0 +1,291 @@
...
+    coefs = np.linspace(0.0,0.05,50)
+
+    fig = make_subplots(rows=1, cols=2, subplot_titles=("Photon", "Pion"))
+
+    # Old convention created the DFs and dicts with the keys 'Photon'/'Pion' instead of 'photons'/'pions'
+    # These should be replaced
+    try:
+        fig.add_trace(go.Scatter(x=coefs, y=effs_by_coef['photons'], name='photons'), row=1, col=1)
+        fig.add_trace(go.Scatter(x=coefs, y=effs_by_coef['pions'], name='pions'), row=1, col=2)
+    except:
```
bfonta, line 280: same as before

bye_splits/scripts/pages/en_page.py
```diff
@@ -0,0 +1,153 @@
+import os
+import sys
+
+#parent_dir = os.path.abspath(__file__ + 4 * '/..')
+parent_dir = '/eos/user/i/iehle/data/PU0/'
```
bfonta, line 5: Can you replace your username by a variable to be passed as an argument? In the future we would like multiple users to be able to launch the web page, if possible.

bye_splits/scripts/pages/en_page.py
```diff
@@ -0,0 +1,153 @@
...
+import bye_splits
+from bye_splits.utils import params, parsing
+
+parser = argparse.ArgumentParser(description='Clustering standalone step.')
+parsing.add_parameters(parser)
+FLAGS = parser.parse_args()
+assert FLAGS.sel in ('splits_only',) or FLAGS.sel.startswith('above_eta_') or FLAGS.sel.startswith('below_eta_')
+
+FLAGS.reg = 'All'
+FLAGS.sel = 'below_eta_2.7'
```
bfonta, line 30: same as before

bye_splits/scripts/pages/en_page.py
```diff
@@ -0,0 +1,153 @@
...
+        df = df.join(max, on='event',rsuffix='_max')
+        df['normed_energies'] = df['en']/df['en_max']
+    elif norm=='PT':
+        df['pt'] = df['en']/np.cosh(df['etanew'])
+        df['normed_energies'] = df['pt']/df['genpart_pt']
+    else:
+        df['normed_energies'] = df['en']/df['genpart_energy']
+
+    df = df[ df['genpart_exeta'] > eta[0] ]
+    df = df[ df['genpart_exeta'] < eta[1] ]
```
bfonta, line 61: same as before

bye_splits/scripts/pages/en_page.py
```diff
@@ -0,0 +1,153 @@
...
+def write_plot_file(input_files, norm, eta, outfile, pars=vars(FLAGS)):
+    normed_energies = dict.fromkeys(input_files.keys(),[0.0]) # Initialize at 0 since we only consider coefs[1:] (coefs[0] is an empty dataframe)
+
+    for key in input_files.keys():
+        
+        if len(input_files[key])==1:
+            with pd.HDFStore(input_files[key][0],'r') as File:
+                coef_strs = File.keys()
+                if norm=='max':
+                    max = File[coef_strs[-1]].set_index('event').drop(columns=['matches','en_max'])
```
bfonta, line 76: Make it somehow clear why it should be the last key to be used.

bye_splits/scripts/pages/en_page.py
```diff
@@ -0,0 +1,153 @@
...
+
+    plot_filename = "{}{}_{}_eta_{}_{}.hdf5".format(parent_dir, normby, plot_file, eta_range[0], eta_range[1])
+
+    if not os.path.exists(plot_filename):
+        normed_dist = write_plot_file(init_files, normby, eta_range, plot_filename)
+    else:
+        try:
+            with pd.HDFStore(plot_filename, "r") as PlotFile:
+                normed_dist = PlotFile['/Normed_Dist'].to_dict(orient='list')
+        except:
```
bfonta, line 125: same as before

bye_splits/scripts/pages/rms_page.py
```diff
@@ -0,0 +1,303 @@
...
+from bye_splits.utils import params, parsing, common
+
+# Set up to work with the rest of the files in the pipeline
+parser = argparse.ArgumentParser(description='Clustering standalone step.')
+parsing.add_parameters(parser)
+FLAGS = parser.parse_args()
+assert FLAGS.sel in ('splits_only',) or FLAGS.sel.startswith('above_eta_') or FLAGS.sel.startswith('below_eta_')
+
+FLAGS.reg = 'All'
+FLAGS.sel = 'below_eta_2.7'
```
bfonta, line 27: same as before

bye_splits/scripts/pages/rms_page.py
```diff
@@ -0,0 +1,303 @@
...
+
+input_files=params.input_files
+
+# Find the element of a list containing strings ['coef_{float_1}', 'coef_{float_2}', ...] which is closest to some float_i
+def_k = 0.0
+def closest(list, k=def_k):
+    try:
+        list = np.reshape(np.asarray(list), 1)
+    except:
```
bfonta, line 36: same as before

bye_splits/scripts/pages/rms_page.py
```diff
@@ -0,0 +1,303 @@
...
+
+    return out
+
+def get_str(coef, file):
+    coef_str = 'coef_{}'.format(str(coef).replace('.','p'))
+    if coef_str not in file.keys():
+        coef_list = [float(re.split('coef_',key)[1].replace('p','.')) for key in file.keys()]
+        new_coef = closest(coef_list, coef)
+        coef_str = 'coef_{}'.format(str(new_coef).replace('.','p'))
+    return coef_str
```
bfonta, line 70: In general all your "pages" seem to have some code duplication? Wouldn't it be better to define common/similar functions in a different script and import that?

bye_splits/scripts/pages/rms_page.py
```diff
@@ -0,0 +1,303 @@
...
+
+def get_rms(coef, eta_range, normby, rms_dict=None, rms_eff_dict=None, binned_rms=True):
+
+    dfs_by_particle = get_dfs(input_files, coef)
+    try:
+        phot_df = dfs_by_particle['photon']
+        pion_df = dfs_by_particle['pion']
+    except:
+        phot_df = dfs_by_particle['photons']
+        pion_df = dfs_by_particle['pions']
```
bfonta, line 116: probably legacy code

bye_splits/scripts/pages/rms_page.py
```diff
@@ -0,0 +1,303 @@
...
+        pion_df['pt'] = pion_df['en']/np.cosh(pion_df['etanew'])
+
+        phot_df['normed_energies'] = phot_df['pt']/phot_df['genpart_pt']
+        pion_df['normed_energies'] = pion_df['pt']/pion_df['genpart_pt']
+
+    phot_df = phot_df[ phot_df['genpart_exeta'] > eta_range[0] ]
+    pion_df = pion_df[ pion_df['genpart_exeta'] > eta_range[0] ]
+
+    phot_df = phot_df[ phot_df['genpart_exeta'] < eta_range[1] ]
+    pion_df = pion_df[ pion_df['genpart_exeta'] < eta_range[1] ]
```
bfonta, line 132: same as before

bye_splits/tasks/cluster.py
```diff
@@ -144,6 +144,20 @@ def cluster(pars, **kw):
                 else:
                     dfout = pd.concat((dfout,cl3d[cl3d_cols+['event']]), axis=0)
 
+            if kw['ForEnergy']:
```
bfonta, line 147: Please choose a better name. What does ```ForEnergy``` do? This is important, since you are changing the reconstruction chain that is supposed to perfectly mimick what is done in CMSSW. Is this change really necessary? Why? Can you do the same outside the chain?

bye_splits/utils/common.py
```diff
@@ -13,6 +13,9 @@
 import numpy as np
 import pandas as pd
 
+import re
```
bfonta, line 16: Are you using this package? If not, please remove this line.

bye_splits/utils/common.py
```diff
@@ -126,3 +129,76 @@ def tc_base_selection(df, region, pos_endcap, range_rz):
...
+        self.file = file
+        self.init_pars = {'opt': copy(pars.opt_kw),
+                          'fill': copy(pars.fill_kw),
+                          'smooth': copy(pars.smooth_kw),
+                          'seed': copy(pars.seed_kw),
+                          'cluster': copy(pars.cluster_kw),
+                          'validation': copy(pars.validation_kw),
+                          'energy': copy(pars.energy_kw)}
+        self.addit = os.path.basename(self.file).replace(".root", "")
+        self.file_addit = lambda x: f"{x}_{self.addit}"
```
bfonta, line 154: use ```.format()```; it is more flexible

bye_splits/utils/common.py
```diff
@@ -126,3 +129,76 @@ def tc_base_selection(df, region, pos_endcap, range_rz):
...
+def transform(nested_list):
+    regular_list=[]
+    for ele in nested_list:
+        if type(ele) is list:
+            regular_list.append(ele)
+        else:
+            regular_list.append([ele])
+    return regular_list
+
+class FileDict:
```
bfonta, line 142: Can you justify why do we need this?
If we really do, you are copying all external ```dict```s, so we end up with memory duplication. We either use what we have or the new method; it is not advantageous to use both.
Also, consider replacing all your class methods by decorators.

bye_splits/utils/params.py
```diff
@@ -9,7 +9,9 @@
 sys.path.insert(0, parent_dir)
 
 import numpy as np
+import pandas as pd
 
+user = 'ehle'
```
bfonta, line 14: same as before

bye_splits/utils/params.py
```diff
@@ -48,6 +50,29 @@
     'Placeholder': np.nan,
 }
 
+if user=='ehle':
```
bfonta, line 53: what happens if a different user tries to run this part of the code?

bye_splits/utils/params.py
```diff
@@ -114,7 +139,10 @@ def set_dictionary(adict):
                   (0.050,)*12 ), # BH
       'CoeffB': 0,
       'MidRadius': 2.3,
-      'PtC3dThreshold': 0.5 }
+      'PtC3dThreshold': 0.5,
+      'ForEnergy': False,
```
bfonta, line 143: please clarify this variable's meaning with a better name

cluster_dashapp_reqs.txt
```diff
@@ -0,0 +1,80 @@
+asttokens==2.2.1
```
bfonta, line 1: Is this file really needed? If so, does it make sense to have it under version control? If so, do you really need all this packages, and can't you find a more condensed alternative?

bfonta (commented):
Thanks for the PR; I can see quite some effort was employed for such a big chunk of code! Some major comments:

- There seems to be code duplication across your "pages"; you might want to rewrite it a bit
- Please put your visualization scripts under ```bye_splits/plot/display_clusters```, or similar

But most importantly:
- I feel that it would be beneficial to wait for #18 to be merged, as your code depends a lot on the reconstruction chain, and it would quickly become broken otherwise.

My suggestion: please implement all the comments that are not related to the reconstruction chain. Once #18 is merged try to rebase your branch on top of ```master``` and fix some possible conflicts.

cluster_dashapp_reqs.txt
isehle: The requirements file might actually be the biggest pain here. I tried to create a far more condensed version, but the problem is that conda then has to infer the required dependencies, and dash is exceptionally finicky when it comes to version and channel conflicts. I created an environment from a much smaller requirement file (18 specifications vs. the current 80) and _almost_ everything worked just as well, but it ended up completely breaking the efficiency page of the dash app, which is what caused the memory overload on llr_ui yesterday. Ideally, we would have a ```.yml``` file for the dash dependencies, since this specifies the channels and which packages are dowloaded through ```pip``` instead of conda, which also seems to make a big difference. I'm not sure if this will work for deployment however because I remember reading that it was specifically looking for a ```.txt``` file, though I can't seem to find where I read that now. In any case it's worth looking into and I would like to have the smallest possible version, but it's trickier to find than it seems given dash's particularities.

bye_splits/plot/display_clusters/pages/rms_page.py
```diff
@@ -0,0 +1,263 @@
...
+        phot_df['normed_energies'] = phot_df['pt']/phot_df['genpart_pt']
+        pion_df['normed_energies'] = pion_df['pt']/pion_df['genpart_pt']
+
+    phot_df = (phot_df[ (phot_df.genpart_exeta > eta_range[0]) & (phot_df.genpart_exeta < eta_range[1]) ])
+    pion_df = (pion_df[ (pion_df.genpart_exeta > eta_range[0]) & (pion_df.genpart_exeta < eta_range[1]) ])
+
+    phot_mean_en = phot_df['normed_energies'].mean()
+    pion_mean_en = pion_df['normed_energies'].mean()
+
+    phot_rms = phot_df['normed_energies'].std()/phot_mean_en
```
bfonta, line 82: regarding the naming, this does not seem a RMS

bye_splits/scripts/cl_size_params.yaml
```diff
@@ -0,0 +1,18 @@
...
+    tree: HGCalTriggerNtuple
+    photons: /data_CMS/cms/ehle/L1HGCAL/photons/sample_skims/skim_photon_0PU_bc_stc_hadd.root
+    electrons: /data_CMS/cms/ehle/L1HGCAL/electrons/sample_skims/skim_electron_0PU_bc_stc_hadd.root
+    pions: /data_CMS/cms/ehle/L1HGCAL/pions/skim_pion_0PU_bc_stc_hadd.root
+
+selection:
+    disconnectedTriggerLayers: [2,4,6,8,10,12,14,16,18,20,22,24,26,28]
+    reachedEE: 2
+    deltarThreshold: 0.05
+    mipThreshold: 0.5
```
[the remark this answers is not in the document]
bfonta: you can add ```deltarThreshold``` there

bye_splits/utils/cl_helpers.py
```diff
@@ -0,0 +1,134 @@
...
+                with pd.HDFStore(path, "r") as File:
+                    if len(File.keys())>0:
+                        if ('photon' in filename) or ('photons' in filename):
+                            output_files['photons'].append(path)
+                        elif ('electron' in filename) or ('electrons' in filename):
+                            output_files['electrons'].append(path)
+                        else:
+                            output_files['pions'].append(path)
+    
+    return output_files
```
bfonta, line 134: please add clear doc-strings to all these methods

bye_splits/utils/parsing.py
```diff
@@ -27,3 +26,5 @@ def add_parameters(parser):
                         choices=('max_energy', 'min_distance'),
                         default='min_distance', type=str,
                         help='Clustering algorithm applied.')
+    parser.add_argument('--user', type=str, help='LXPlus username.')
```
bfonta, line 29: LXPlus -> lxplus

bye_splits/utils/parsing.py
```diff
@@ -27,3 +26,5 @@ def add_parameters(parser):
...
+    parser.add_argument('--cluster_studies', type=bool, help='boolean, whether or not to read and write files for the cluster size studies steps.', default=False)
```
bfonta, line 30: long help string, please indent and make it more concise; you can also remove "boolean", sinice it is clear from its type

run_cluster_app.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/bash
```
bfonta, line 1: use ```#!/usr/bin/env bash``` [source](https://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang)

run_cluster_app.sh
```diff
@@ -0,0 +1,5 @@
...
+
+user=$1
+
+python bye_splits/plot/display_clusters/cl_app.py --sel below_eta_2.7 --reg All --user "$user"
```
bfonta, line 5: you really want eta < 2.7?

bfonta: You can use ```--user $1``` directly

README.org
```diff
@@ -128,6 +172,17 @@ We are currently running a pod at https://viz2-hgcal-event-display.app.cern.ch/.
 + [[https://cloud.google.com/kubernetes-engine/docs/concepts/pod][What is a pod]]?
 + [[https://paas.docs.cern.ch/3._Storage/eos/][How to mount =/eos= at CERN so that it is accessible by a pod?]]
 
+*** DashApp For Cluster Radii Studies
+A DashApp has been built to interactively explore the effect of cluster size on various cluster properties, which is currently hosted at https://bye-splits-app-hgcal-cl-size-studies.app.cern.ch/.
+To run the app locally, you can do:
+
+#+BEGIN_SRC shell
+bash run_cluster_app.sh <username>
```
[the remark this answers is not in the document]
isehle: This is already explained in the README on [this line](https://github.com/isehle/bye_splits/blob/44d5160dd41656c012f66217a8f75977e9f010ea/README.org?plain=1#L185); this will work with any username provided they have access to my directory as well (which both you and @mchiusi do), because it first searches for the files in your directory before searching for them in mine. Any output files will be written to your directory.

README.org
```diff
@@ -128,6 +172,17 @@ We are currently running a pod at https://viz2-hgcal-event-display.app.cern.ch/.
...
+#+END_SRC
+
+where =<username>= is your lxplus username. The app reads the configuration file =bye_splits/plot/display_clusters/config.yaml= and assumes that you have a directory structure equivalent to the =eos/= directory described in the cluster size step.
+
+If you have run the cluster size study yourself, the generated files that the Dash App takes as input will be under either the =PU0/= or =PU200/=, depending on the boolean 'PileUp' in the configuration file. It then performs the necessary analysis on these files to generate the data for each page, which are themselves written to files in this directory. In order to minimize duplication and greatly speed up the user experience, if one of these files does not exist in your own directory, it looks for it under =/eos/user/i/iehle/data/<pile_up>= where a large number of the possible files already exist. The same procedure is used for reading the generated cluster size files, so you can use the app without having had to run the study yourself.
```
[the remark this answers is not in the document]
isehle: Yup, I can definitely move the path to the table and link back to it.

bye_splits/data_handle/base.py
```diff
@@ -52,8 +52,8 @@ def tree_path(self):
         
 class BaseData(abc.ABC):
     """Base data management class."""
-    def __init__(self, inname, tag, reprocess, logger, is_tc):
-        with open(params.CfgPaths['prod'], 'r') as afile:
+    def __init__(self, prod_key, inname, tag, reprocess, logger, is_tc):
+        with open(params.CfgPaths[prod_key], 'r') as afile:
```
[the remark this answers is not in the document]
isehle: It's used when calling ```get_data_reco_chain_start()```, for example [here](https://github.com/isehle/bye_splits/blob/44d5160dd41656c012f66217a8f75977e9f010ea/bye_splits/scripts/cluster_size.py#L44). This allows me to specify a configuration file when reading the data instead of it being hardcoded to read ```prod_params.yaml```, which is how it's currently written. Furthermore, even if I did use the default configuration file, I think it makes the code easier to read when you see which file is being used at this point, instead of having to hunt for it in the ```data_handle/``` scripts.

bye_splits/data_handle/event.py
```diff
@@ -146,15 +148,15 @@ def provide_random_events(self, n, seed=None):
         """Provide 'n' random events."""
         if seed is not None:
             self.rng = np.random.default_rng(seed=seed)
-        events = self.rng.choice(self.ev_numbers, size=n, replace=False)
+        events = self.rng.choice(self.ev_numbers, size=n, replace=False) if n!=-1 else -1
```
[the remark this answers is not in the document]
isehle: In fact I based this change on #26, copying the line [here](https://github.com/bfonta/bye_splits/blob/1a0cd5030ad3ff536da2c0e3c480d6d5776b4f83/bye_splits/data_handle/event.py#L149). Do you mean that you plan on changing it further? In any case, I don't imagine it will cause any large problems.

bye_splits/data_handle/event.py
```diff
@@ -146,15 +148,15 @@ def provide_random_events(self, n, seed=None):
...
         return self.provide_events(events), events
         
-    def select(self):
+    def select(self):        
         with up.open(self.indata.path, array_cache='550 MB', num_workers=8) as f:
             tree = f[self.indata.tree_path]
             allvars = set([y for x in self.var.values() for y in x.values()])
             data = tree.arrays(filter_name='/' + '|'.join(allvars) + '/',
-                               library='ak')
+                               library='ak', entry_stop=10)
```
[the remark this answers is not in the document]
isehle: Yup you're totally right, sorry about that, should've caught that before pushing.

bye_splits/plot/display_clusters/cl_app.py
```diff
@@ -0,0 +1,61 @@
...
+import dash
+
+import dash_bootstrap_components as dbc
+from dash_bootstrap_templates import load_figure_template
+
+parent_dir = os.path.abspath(__file__ + 4 * '/..')
+sys.path.insert(0, parent_dir)
+
+app = Dash(__name__, use_pages=True, external_stylesheets=[dbc.themes.SOLAR])
+load_figure_template('SOLAR')
```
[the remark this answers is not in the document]
isehle: I agree it'd be best to have one consistent theme, but perhaps we can talk about this in person (personally I prefer SOLAR but I'm open to changing it if FLATLY is better for @mchiusi).

bye_splits/plot/display_clusters/config.yaml
```diff
@@ -0,0 +1,12 @@
+dirs:
+  Local: False
+  LocalDir: '/home/llr/cms/ehle/NewRepos/bye_splits/data/new_algos/'
+  EhleDir: '/eos/user/i/iehle/'
```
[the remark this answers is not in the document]
isehle: I can change the names if you'd like but let me know if you have any suggestions, because "local" and "localDir" sound fairly generic to me as is.
The reason I have "ehleDir" specified specifically here is because the script will always check my directory for input files if it can't find them in ```user```'s directory, since writing the files can take some time so I want to reduce the as much duplication as possible.

bye_splits/plot/display_clusters/pages/eff_page.py
```diff
@@ -0,0 +1,230 @@
...
+
+# Takes a dataframe 'df' with a column 'norm' to normalize by, and returns
+# 1) a binned matching efficiency list
+# 2) a binned list corresponding to 'norm' 
+# where the binning is done by percentage 'perc' of the size of the 'norm' column
+def binned_effs(df, norm, perc=0.1):
+    eff_list = [0]
+    en_list = [0]
+    en_bin_size = perc*(df[norm].max() - df[norm].min())
+    if not perc==1.:
```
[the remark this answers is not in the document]
isehle: Good catch(es)!

bye_splits/plot/display_clusters/pages/eff_page.py
```diff
@@ -0,0 +1,230 @@
...
+
+    fig.update_yaxes(type="log")
+
+    fig.update_layout(title_text="Efficiency/{}".format(normby), yaxis_title_text=r'$Eff (\frac{N_{Cl}}{N_{Gen}})$')
+
+    return fig, dbc.Table.from_dataframe(glob_effs)
+
+def write_eff_file(norm, coefs, eta, file):    
+    effs_dict = {'photons': [0.0],
+                 'pions': [0.0]}
```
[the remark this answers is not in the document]
isehle: It's actually not a placeholder; I should have wrote a comment, adding it now. The point is that we don't have a DataFrame for ```coef==0``` (or more specifically, we have an empty DataFrame), so we can't calculate efficiencies there. But a cluster of radius zero has efficiency of zero, so we initialize the list as such.

bye_splits/plot/display_clusters/pages/eff_page.py
```diff
@@ -0,0 +1,230 @@
...
+    for coef in coefs[1:]:
```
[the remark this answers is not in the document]
isehle: Explained in the same comment that I added to explain why we initialize the lists at ```[0.0]```.

bye_splits/plot/display_clusters/pages/en_page.py
```diff
@@ -0,0 +1,186 @@
...
+    html.P("Normalization:"),
+    dcc.Dropdown(['Energy', 'PT'], 'PT', id='normby')
+])
+
+def fill_dict_w_mean_norm(key, coef, eta, df, norm, out_dict, df_needs_norm=False):
+    """The df_needs_norm bool handles legacy code which calculated the norm in this file,
+    while the updated version of the cluster size study calculates it directly."""
+    if df_needs_norm:
+        if norm=='PT':
+            df['pt'] = df['en']/np.cosh(df['etanew'])
```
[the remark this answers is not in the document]
isehle: Yup, good catch, thanks!

bye_splits/plot/display_clusters/pages/en_page.py
```diff
@@ -0,0 +1,186 @@
...
+            df['normed_energies'] = df['pt']/df['genpart_pt'] if 'etanew' in df.keys() else df['en']/np.cosh(df['eta'])
+        else:
+            df['normed_energies'] = df['en']/df['genpart_energy']
+
+        df = df[ (df.genpart_exeta > eta[0]) & (df.genpart_exeta < eta[1]) ]
+
+        mean_energy = df['normed_energies'].mean()
+    else:
+        df = df[ (df.gen_eta > eta[0]) & (df.gen_eta < eta[1]) ]
```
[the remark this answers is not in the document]
isehle: Yes it's correct; the boolean ```df_needs_norm``` which the if/else statement checks is there to handle legacy code where the columns were named differently. Furthermore, in the old code, the normalization was done in ```en_page.py```, whereas now I add normalized columns when first creating the files in ```cluster_size.py```. So the switch handles these two cases.

bye_splits/plot/display_clusters/pages/en_page.py
```diff
@@ -0,0 +1,186 @@
...
+                    except KeyError:
+                        pass
+                    fill_dict_w_mean_norm(key, coef, eta, df, norm, normed_energies, df_needs_norm=True)
+        else:
+            file_list = [pd.HDFStore(val,'r') for val in input_files[key]]
+            coef_strs = file_list[0].keys()
+            for coef in coef_strs[1:]:
+                df_list = [file_list[i][coef] for i in range(len(file_list))]
+                full_df = pd.concat(df_list)
+                full_df = full_df.set_index('event') if not full_df.index.name=='event' else full_df
```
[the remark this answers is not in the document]
isehle: I agree, but I get a syntax error if I don't include it, so I can keep it like this or use a traditional if statement. Let me know what you prefer.

bye_splits/plot/display_clusters/pages/en_page.py
```diff
@@ -0,0 +1,186 @@
...
+        if len(input_files[key])==0:
+            continue
+        elif len(input_files[key])==1:
+            with pd.HDFStore(input_files[key][0],'r') as File:
+                coef_strs = File.keys()
+                for coef in coef_strs[1:]:
+                    df = File[coef].set_index('event') if not File[coef].index.name=='event' else File[coef]
+                    try:
+                        df = df.drop(columns=['matches', 'en_max'])
+                    except KeyError:
```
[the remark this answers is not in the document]
isehle: This also handles legacy code; older versions had the two columns, the new versions (since we introduced the skimming step—I can't find the specific PR that implemented this) do not include them.

bye_splits/plot/display_clusters/pages/en_page.py
```diff
@@ -0,0 +1,186 @@
...
+
+    if normby=='Energy':
+        y_axis_title = r'$\huge{\frac{\bar{E_{Cl}}}{\bar{E}_{Gen}}}$'
+    elif normby=='PT':
+        y_axis_title = r'$\huge{\frac{\bar{p_T}^{Cl}}{\bar{p_T}^{Gen}}}$'
+    else:
+        y_axis_title = r'$\huge{\frac{E_{Cl}}{E_{Max}}}$'
+
+    plot_filename = "{}_{}_eta_{}_{}.hdf5".format(normby, plot_file, eta_range[0], eta_range[1])
+    plot_filename_user = "{}{}".format(f"{data_dir}", plot_filename)
```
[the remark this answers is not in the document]
isehle: Can't imagine what I was thinking there.

bye_splits/plot/display_clusters/pages/en_page.py
```diff
@@ -0,0 +1,186 @@
...
+
+    coef_labels = [round(coef,3) for coef in coefs]
+    coef_labels= coef_labels[0::5]
+
+    fig = make_subplots(rows=1, cols=2, subplot_titles=("Photon", "Pion"))
+    
+    try:
+        en_phot = normed_dist['photon']
+        en_pion = normed_dist['pion']
+    except:
```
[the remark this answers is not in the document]
isehle: I agree, but it means updating the key names in all of the old files. It shouldn't take long but I'd like to keep the try-except until I'm sure all of the files have been updated.

bye_splits/plot/display_clusters/pages/pile_up_page.py
```diff
@@ -0,0 +1,190 @@
...
+    # Cluster DFs
+    phot_cl3d_df, el_cl3d_df = df_dict['photons']['cl3d'], df_dict['electrons']['cl3d']
+
+    # Apply angular restrictions
+    phot_cl3d_df = phot_cl3d_df[ (phot_cl3d_df.good_cl3d_eta > eta_range[0]) & (phot_cl3d_df.good_cl3d_eta < eta_range[1]) ]
+    el_cl3d_df = el_cl3d_df[ (el_cl3d_df.good_cl3d_eta > eta_range[0]) & (el_cl3d_df.good_cl3d_eta < eta_range[1]) ]
+
+    # Apply energy and pt thresholds
+    phot_cl3d_df = phot_cl3d_df[ (phot_cl3d_df.good_cl3d_pt > min_pt) & (phot_cl3d_df.good_cl3d_energy > min_en) ]
+    el_cl3d_df = el_cl3d_df[ (el_cl3d_df.good_cl3d_pt > min_pt) & (el_cl3d_df.good_cl3d_energy > min_en) ]
```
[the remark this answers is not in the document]
isehle: It's correct; the thresholds are set by the user in the app, so even though a typical use-case would be to apply a pT cut and leave the energy threshold at 0, this allows for checking a combination if one would want to.
Also, at the moment these thresholds are defined as cuts on the clusters themselves, but once I'm able to have more TC information, I'd like to change these to thresholds on TCs; unclear how much work that's going to entail, since the matching and clustering is done earlier.

bye_splits/plot/display_clusters/pages/rms_page.py
```diff
@@ -0,0 +1,263 @@
...
+    sum_of_squares = 0
+    for i in data:
+        sum_of_squares+=i**2
+    return np.sqrt(sum_of_squares/size)
+
+def effrms(data, c=0.68):
+    """Compute half-width of the shortest interval
+    containing a fraction 'c' of items in a 1D array.
+    """
+    out = {}
```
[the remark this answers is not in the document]
isehle: Yes, I think this is left over from an older version. Thanks!

bye_splits/plot/display_clusters/pages/rms_page.py
```diff
@@ -0,0 +1,263 @@
...
+    new_series = data.dropna()
+    x = np.sort(new_series, kind='mergesort')
+    m = int(c *len(x)) + 1
+    out = [np.min(x[m:] - x[:-m]) / 2.0]
```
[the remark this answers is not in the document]
isehle: Yes, ```data``` must be a 1D list/array. Let me know if this helps at all:
effRMS

bye_splits/plot/display_clusters/pages/rms_page.py
isehle: Worth discussing further, because I've also been confused by this. I originally calculated this using the direct formula for RMS (can find it [here](https://encyclopedia.thefreedictionary.com/Root+mean+square) for example), though JB advised me to do this instead. My (vague) understanding is that while this might not be exactly what a mathematician calls an RMS, it's what we call RMS, and is essentially just a normalized and shifted version of RMS (note that standard deviation and RMS are equivalent when the mean is zero).

bye_splits/tasks/cluster.py
```diff
@@ -131,6 +133,12 @@ def cluster(pars, **kw):
             sout['data'] = dfout
 
         nevents = dfout.event.unique().shape[0]
+        
+        if kw['GenerateClusterSize']:
+            coef = 'coef_'+str(kw['CoeffA'][0]).replace('.','p')
+            with pd.HDFStore(kw['ClusterSizePath'], mode='a') as clSizeOut:
+                clSizeOut[coef] = dfout
+
```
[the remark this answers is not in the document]
isehle: I don't think I understand; this doesn't change any names. If ```ClusterSizePath``` is set to ```False```, ```cluster.py``` runs completely equivalently to how it did before. If it's set to true, then a _new_ file is created whose path is stored in ```kw['ClusterSizePath']``` (which is filled in ```cluster_size.py``` and is not used anywhere else in the code).

run_cluster_app.sh
isehle: Yes we were originally interested in just eta < 2.7, though going forward I think it's best to keep 'all' and the user can apply the desired cuts in the app.

For the user option, yes I should change it; I previously had a few arguments and it made readability easier, though now that I've trimmed it down to one I agree it's too redundant.

bye_splits/plot/display_clusters/pages/eff_page.py
```diff
@@ -0,0 +1,230 @@
...
+                              }, index=[0])
+
+    fig = make_subplots(rows=1, cols=2, subplot_titles=("Photon", "Pion"))
+
+    fig.add_trace(go.Scatter(x=phot_x, y=phot_effs, name='photons'), row=1, col=1)
+    fig.add_trace(go.Scatter(x=pion_x, y=pion_effs, name='pions'), row=1, col=2)
+
+    fig.update_xaxes(title_text="{} (GeV)".format(normby))
+
+    fig.update_yaxes(type="log")
```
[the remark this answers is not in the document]
isehle: The quickest and simplest way for me to make these changes was to use a Python autoformatter (I'm using [Black](https://github.com/psf/black)), but this necessarily reformatted the code in many more ways. Let me know if you're okay with their current [style guide](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html), and if not I think we should all agree on one formatter+configuration to use.

bye_splits/plot/display_clusters/pages/pile_up_page.py
```diff
@@ -0,0 +1,190 @@
...
+                else:
+                    cl3d_df = pd.read_parquet(cl3d_outfile)
+
+                df_dict[key] = {'genpart': genpart_df, 'cl3d': cl3d_df}
+
+    return df_dict
+
+df_dict = generate_dfs(input_files)
+
+def create_figures(phot_df, el_df, ptype, hist_bins):
```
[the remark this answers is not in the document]
isehle: Handled by [Black](https://github.com/psf/black).

bye_splits/scripts/cl_size_params.yaml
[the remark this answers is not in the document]
isehle: Though I understand this duplicates the code in ```prod_params.yaml```, it's much simpler to have one configuration file that contains the specific information that ```cluster_size.py``` requires. While it is of course possible to have a separate configuration file which contains only the specific information under ```cl_size``` (i.e., [here](https://github.com/isehle/bye_splits/blob/44d5160dd41656c012f66217a8f75977e9f010ea/bye_splits/scripts/cl_size_params.yaml#L14)) and to combine that with the information in ```prod_params.yaml```, this feels like it overcomplicates things unnecessarily simply to avoid duplicating a few lines of code.

bye_splits/scripts/cl_size_params.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+cl_size:
+    particles: 'photons'
+    reinit: False
+    ClusterSizeBaseName: 'cluster_size_pileup'
+    Coeffs: [0.0, 0.05, 50]
```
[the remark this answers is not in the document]
isehle: What exactly do you mean by "my" folder? I suppose you mean the ```bye_splits/plot/display_clusters```, but this configuration has nothing to do with the plotting, hence why I have it together with ```cluster_size.py``` under ```bye_splits/scripts```. Furthermore, though I know this is likely to change in the future, currently I am the only one who has any files at all under ```bye_splits/scripts```. If you would like, I can create a subfolder ```bye_splits/scripts/cluster_size``` and keep both files in there, but I don't want to move these to the plot folder.

bye_splits/tasks/config.yaml
```diff
@@ -49,6 +49,8 @@ cluster:
     CoeffB: 0
     MidRadius: 2.3
     PtC3dThreshold: 0.5
+    GenerateClusterSize: False
+    ClusterSizePath: None
```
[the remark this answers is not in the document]
isehle: ```clusterSizePath``` can be removed since the configuration dictionary can be appended to in ```cluster_size.py```, but ```generateClusterSize``` is checked in ```cluster.py``` to determine whether or not to create/append to a new ```.hdf5``` file. If it's removed, anyone who wasn't running the cluster size study would receive a ```KeyError``` on [this line](https://github.com/isehle/bye_splits/blob/44d5160dd41656c012f66217a8f75977e9f010ea/bye_splits/tasks/cluster.py#L137).
If you really don't want this in the configuration file, we can change  that line in ```cluster.py``` to something like ```if "generateClusterSize" in kw.keys()```, but personally, I think it's clearer to keep the True/False switch in the configuration file to improve readability in ```cluster.py```.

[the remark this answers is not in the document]
isehle: To be clear, its value would always be kept as ```False``` in the YAML file in order not to interfere with any other scripts/configurations. The cluster size script simply switches the value in the loaded dictionary before passing it to ```cluster.py```.

README.org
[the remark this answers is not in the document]
isehle: Currently having trouble linking to the table; currently the link seems to just reload the page, at least in [this gist](https://gist.github.com/isehle/9e088f6817e5f740ccb14132036be86c#file-testing-org). This isn't the full README, but I tried to include enough of it that the link at the bottom would bring us back to the top, but it doesn't. Any advice?

bye_splits/plot/display_clusters/pages/eff_page.py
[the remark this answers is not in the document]
bfonta: Ok. It was not clear that the coefficient was the radius; please mention it in the comment you are adding.

bye_splits/plot/display_clusters/pages/en_page.py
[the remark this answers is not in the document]
bfonta: I think you should not worry with legacy code, as the repository should be up-to-date; this would also avoid cluttering the code with many ```if/else``` conditions. Let me know if you find more places where names changed.

bye_splits/plot/display_clusters/pages/en_page.py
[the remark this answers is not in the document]
bfonta: I suppose the traditional is better here; you can also keep it in one line :)

bye_splits/plot/display_clusters/pages/en_page.py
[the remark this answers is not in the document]
bfonta: same answer as above

bye_splits/plot/display_clusters/pages/en_page.py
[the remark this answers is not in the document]
bfonta: same as above; if you keep this we will likely forget to change it later on; can't you do it only locally if needed? I am just trying to make the code as simple as possible for future users.

bye_splits/plot/display_clusters/pages/pile_up_page.py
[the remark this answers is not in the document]
bfonta: Fine, and useful; just make sure the default cut on the energy is zero.

bye_splits/plot/display_clusters/pages/rms_page.py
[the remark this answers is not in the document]
bfonta: I got it now, nice algorithm. Thank you for the diagram.

- could you ```assert``` the shape of the input for extra clarity?
- why do you convert the output to a ```list()```? If needed later outside the function, please do it where needed.

bye_splits/plot/display_clusters/pages/rms_page.py
bfonta: We usually call it the detector *response* when dividing some spread-like measure by the energy mean. The former could be the RMS, the effective RMS, or the standard deviation.

bye_splits/scripts/cl_size_params.yaml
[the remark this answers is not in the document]
bfonta: Ok.

bye_splits/scripts/cl_size_params.yaml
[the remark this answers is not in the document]
bfonta: Ok, then let us do as follows:

- keep your script under ```bye_splits/scripts```
- move all python scripts under ```bye_splits/``` to the new ```bye_splits/scripts``` folder, including ```__init__.py```, and make sure you can run them via ```python -m bye_splits.scripts.<name_of_script>``` (testing your script and ```run_chain.py``` is enough, as the others will be either deleted or updated in the future).

bye_splits/tasks/cluster.py
```diff
@@ -29,7 +31,7 @@ def cluster(pars, **kw):
         radiusCoeffB = kw['CoeffB']
         empty_seeds = 0
 
-        for tck, seedk in zip(tc_keys, seed_keys):
+        for tck, seedk in tqdm(zip(tc_keys, seed_keys), total=len(tc_keys)):
             tc = sin_tc[tck]
```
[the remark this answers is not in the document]
bfonta: Yes, adding it outside the chain is fine.

bye_splits/tasks/config.yaml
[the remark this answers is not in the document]
bfonta: same as previous comment

bye_splits/plot/display_clusters/pages/en_page.py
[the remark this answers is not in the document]
isehle: These try/except statements (specifically those handling the singular/plural switch) are now handled by a function, so any user reading the code will notice that the dataframes are received from the dictionaries using this function. That being said, I agree that it's a bit of a forced fix, and if possible I'd like to just rename the old keys to match the new ones.

bye_splits/plot/display_clusters/pages/rms_page.py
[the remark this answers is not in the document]
isehle: Thanks! And yup, the assert statement has been added. It's unclear why I had originally converted to a list, it seems I ended up converting it back to a float later in the code—fixed now.

bye_splits/plot/display_clusters/pages/rms_page.py
```diff
@@ -0,0 +1,263 @@
...
+        
+    nice_coefs = np.linspace(0.0,0.05,50)
+
+    fig = make_subplots(rows=1, cols=2, subplot_titles=("Photon", "Pion"))
+
+    fig.add_trace(go.Scatter(x=nice_coefs, y=rms_by_part['Photon'], name='RMS', line_color='purple'), row=1, col=1)
+    fig.add_trace(go.Scatter(x=nice_coefs, y=rms_eff_by_part['Photon'], name='Eff-RMS', line_color='purple', mode='markers'), row=1, col=1)
+    
+    fig.add_trace(go.Scatter(x=nice_coefs, y=rms_by_part['Pion'], name='RMS', line_color='red'), row=1, col=2)
+    fig.add_trace(go.Scatter(x=nice_coefs, y=rms_eff_by_part['Pion'], name='Eff-RMS', line_color='red', mode='markers'), row=1, col=2)
```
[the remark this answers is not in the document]
isehle: Done!

bye_splits/data_handle/event.py
```diff
@@ -19,9 +19,9 @@
 from data_handle.base import BaseData
 
 class EventData(BaseData):
-    def __init__(self, inname, tag='v0', default_events=[],
+    def __init__(self, prod_key, inname, tag='v0', default_events=[],
```
bfonta, line 22: Wasn't this supposed to be removed?

isehle: No—it's significantly more convenient to allow the user to pass in a specified configuration file than having it be hardcoded to use a specific one. We can however add a default option if you would like.

bye_splits/production/prod_params.yaml
```diff
@@ -6,6 +6,7 @@ io:
     pions: /data_CMS/cms/alves/L1HGCAL/skim_small_TEST_module_TEST_BIG.root
 
 selection:
+    particles: photons
```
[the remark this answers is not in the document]
isehle: The command line argument was used specifically for the skimming step; the reco chain itself has never had this option, and as you use this configuration file to load the data, it seems a good place to keep the information. Specifically since this contains other necessary input information but you had had it hardcoded to read only "photons" before, leaving the "electrons" and "pions" paths completely untouched unless someone went in to change [this line](https://github.com/bfonta/bye_splits/blob/e509e31d9252d1a372ca9e054d61af250c35c88a/bye_splits/data_handle/data_handle.py#L20).

run_cluster_size.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-python bye_splits/scripts/cluster_size.py --reg All --user "$1" --cluster_studies True
\ No newline at end of file
+python bye_splits/scripts/cluster_size/run_cluster_size.py --reg All --user "$1" --cluster_studies True
```
bfonta, line 3: You could use ```store_true```; the argument ```True``` for a flag is redundant.

bye_splits/tasks/cluster.py
```diff
@@ -142,9 +142,7 @@ def cluster(pars, **kw):
         )
 
     if dfout is not None:
-        out = common.fill_path(
-            kw["ClusterOutPlot"], data_dir="data/new_algos/electrons/radii", **pars
-        )
+        out = common.fill_path(kw["ClusterOutPlot"], **pars)
```
bfonta, line 145: Is "Black" changing the syntax it changed before?

bfonta (approved):

bye_splits/tasks/cluster.py
isehle: Yes it depends on how long the line (without passing ```data_dir``` it goes to one line since it considers it short enough).

mchiusi (approved):

bfonta:
I've now realized why converting ```particles``` to a configuration parameter is problematic: it does not allow instantiating more than one particle type at once, for instance [here](https://github.com/bfonta/bye_splits/blob/e509e31d9252d1a372ca9e054d61af250c35c88a/bye_splits/plot/display/main.py#L43-L46). Since the latter is essential for any event display, #26 will introduce the parameter back to the code, keeping the one from the configuration as a default when the user does not specify it directly.

bfonta:
@isehle Another point: you mention a configuration file [here](https://github.com/bfonta/bye_splits/blob/ff9e6ec554c89ef70038cc4d48460d960bf3f76e/bye_splits/utils/params.py#L34-L35) that was not included in the PR.
Instead of fixing this, I will rewrite the way we are getting the configuration, to simplify it: a single file in the top-level directory, as discussed today. This will also be included in #26.

PR #30, opened 2023-03-08
add ROI finder algorithm and minor changes in the display code

mchiusi:
All the changes are at _display level_: they do not affect the back-end code which provide geometry and events.

1) The **ROI finder algorithm** (written [here](https://github.com/mchiusi/bye_splits/blob/main_correct/bye_splits/plot/display_plotly/event_processing.py#L93)) can be divided in three parts:
  - Choose k-layers (by default k = 4) window based on energy deposited in each layer;
  - Choose the (u,v) coordinates of the module corresponding to max dep-energy;
  - Extend the selection to the nearby modules that have at least 30% of the energy in the selected module and at least 10 mipT od deposited energy.

_Caveat_: this simple algorithm defines one ROI per event. For the moment, since we are using single particles, it works fine (in both 0PU and 200PU events).

2) A new **dashApp theme** (called 'FLATLY') has been added.

3) The generated **particle information** has been added on screen:
 - extrapolated eta coordinate;
 - extrapolated phi coordinate;
 - energy in [GeV].


bye_splits/data_handle/config.yaml
```diff
@@ -27,7 +27,8 @@ varEvents:
     eta: 'good_tc_eta'
     idcl: 'good_tc_cluster_id'
     pt: 'good_tc_pt'
-    en: 'good_tc_mipPt'
+    mip: 'good_tc_mipPt'
+    en: 'good_tc_energy'
```
bfonta, line 31: This is dangerous, as these variables are being used elsewhere. I have to ask you to run both the 2D display and the reconstruction chain to make sure they work.

bfonta: For the 2D display you can launch:

```python
python bye_splits/plot/display/main.py
```

and for the reco chain just run the ```bye_splits/reco_chain.py``` example.

bye_splits/plot/display_plotly/event_processing.py
```diff
@@ -69,14 +89,42 @@ def update_3dfigure(fig, df):
...
+    ''' Choose k-layers window based on energy deposited in each layer '''
+    layer_sums = df.groupby(['layer']).mipPt.sum() 
+    initial_layer = (layer_sums.rolling(window=k).sum().shift(-k+1)).idxmax()
+    
+    mask = (df.layer>=initial_layer) & (df.layer<(initial_layer+2*k))
+    input_df = df[mask]
+    
+    ''' Choose the (u,v) coordinates of the module corresponding to max dep-energy.
+    Extend the selection to the nearby modules with at least 30% max energy and 10 mipT. '''
+    module_sums = input_df.groupby(['waferu','waferv']).mipPt.sum()
```
bfonta, line 102: Aren't you also grouping modules belonging to different layers?

bye_splits/plot/display_plotly/event_processing.py
```diff
@@ -88,51 +136,45 @@ def update_2dfigure(fig, df):
...
-        html.Button(children='Random event', id='event-val', n_clicks=0),
-        html.Button(children='Submit selected event', id='submit-val', n_clicks=0),
+        dbc.Button(children='Random event', id='event-val', n_clicks=0),
+        dbc.Button(children='Submit selected event', id='submit-val', n_clicks=0),
         html.Div(id='event-display', style={'display':'inline-block', "margin-left": "15px"}),
         html.Div(id='which', style={'display':'inline-block', "margin-left": "15px"}),
     ]),
     dcc.Graph(id='graph2d'),
-    dcc.Store(id='dataframe')])
+    dcc.Store(id='dataframe')]), ])
```
bfonta, line 180: In this file you could use the ```**options``` syntax to avoid repeating the same arguments.

bye_splits/plot/display_plotly/main.py
```diff
@@ -10,73 +10,95 @@
...
         df_cluster    = df_sel[df_sel.tc_cluster_id != 0]
         fig = processing.set_3dfigure(df_cluster)
         fig = processing.update_3dfigure(fig, df_no_cluster)
     else:
         fig = processing.set_3dfigure(df_sel)
-    
-    if layer == 'display the entire event':
+   
+    if 'ROI' in checkbox:
+        fig = processing.add_ROI(fig, df_sel)
```
bfonta, line 90: If I understand correctly, there is no need to return ```fig```, and ```add_ROI()``` does not need a return value.

bye_splits/plot/display_plotly/main.py
```diff
@@ -92,7 +114,6 @@ def make_graph(data, slider_value, cluster, mip, page):
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
-    parser.add_argument('-id','--username',type=str,default=os.getlogin())
     parser.add_argument('--host',type=str,default='llruicms01.in2p3.fr')
     parser.add_argument('--port',type=int,default=8004)
```
bfonta, line 118: Can you add short help strings to each option?

bye_splits/plot/display_plotly/plot_event.py
```diff
@@ -45,3 +50,17 @@ def produce_2dplot(df, opacity=1):
...
     return listdata
+
+
+def plot_modules(df):
+    array_data = df[['hex_x','hex_y','z']].to_numpy()
+    listdata = []
+    for j,i in enumerate(array_data):
+        x1 = np.append(i[0],i[0][0])
+        y1 = np.append(i[1],i[1][0])
+        z1 = np.array(int(len(x1)) * [i[2]])
```
bfonta, line 61: ```np.full_like()``` ([here](https://numpy.org/doc/stable/reference/generated/numpy.full_like.html)) would be more idiomatic, and very likely faster.

bye_splits/utils/params.py
```diff
@@ -17,8 +17,8 @@ def read_task_params(task):
         d.update(cfg[task])
     return d
 
-LocalStorage = os.path.join(parent_dir, 'data/new_algos')
-EOSStorage = lambda u: os.path.join('/eos/user', u[0], u, 'FPGAs/new_algos')
+LocalStorage = os.path.join(parent_dir, 'data/new_algos/')
+EOSStorage = lambda u: os.path.join('/eos/user', u[0], u, 'visualization/display3d/')
```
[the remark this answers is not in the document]
isehle: In my code I updated the lambda function to accept the trailing directory as a second parameter, see [here](https://github.com/isehle/bye_splits/blob/d3fca8462dc833fea9ea71d04c98fc04386fc83a/bye_splits/utils/params.py#L24). This is mentioned in the PR as something that can break the code.

bfonta:
Additionally, could you please explain how to run the 3D event display in [this section](https://github.com/bfonta/bye_splits/tree/e509e31d9252d1a372ca9e054d61af250c35c88a#event-visualization) of the README?

bye_splits/data_handle/config.yaml
mchiusi: For the moment I can just leave the previous name and add the `en_gev` as key for the branch called `good_tc_energy`.

bye_splits/plot/display_plotly/event_processing.py
mchiusi: Yes. In the following way:
 - `input_df` ([here](https://github.com/mchiusi/bye_splits/blob/main_correct/bye_splits/plot/display_plotly/event_processing.py#L98)) contains all the TCs involved in the event selected and belonging to the selected k-layers.
 - thanks to [this](https://github.com/mchiusi/bye_splits/blob/main_correct/bye_splits/plot/display_plotly/event_processing.py#L102) command I can easily group-by TCs belonging to different layers having the same (u, v) coordinate
As a result, I'm grouping modules belonging to different layers.

I can explicit it a bit more in the caption.

bye_splits/plot/display_plotly/event_processing.py
mchiusi: Done it! Thanks for your suggestion. I merged two callbacks in one, too.

bye_splits/plot/display_plotly/main.py
mchiusi: Sure

bye_splits/plot/display_plotly/event_processing.py
bfonta: My point was that it seems you are calculating the sums across multiple layers, when by definition a module sum refers to a single layer only. 
However, I now understand that your algorithm does not actually require calculating individual module sums. But, as discussed today, this will have to be rewritten as an independent scripts under ```task```, and there individual modules will be necessary.

bye_splits/plot/display_plotly/event_processing.py
```diff
@@ -69,14 +89,43 @@ def update_3dfigure(fig, df):
         fig.add_trace(list_scatter[index])
     return fig
 
+def add_ROI(fig, df, k=4):
+    ''' Choose k-layers window based on energy deposited in each layer '''
+    layer_sums = df.groupby(['layer']).mipPt.sum() 
+    initial_layer = (layer_sums.rolling(window=k).sum().shift(-k+1)).idxmax()
+    
+    mask = (df.layer>=initial_layer) & (df.layer<(initial_layer+2*k))
```
bfonta, line 97: You are using the fact that only one each two layers have trigger cells. This has to be changed for two reasons:
- there is no guarantee this will be the case (although very likely): you should use the ```disconnectedTriggerLayers``` parameters instead
- this is not true for the CE-H

bye_splits/plot/display_plotly/event_processing.py
```diff
@@ -69,14 +89,43 @@ def update_3dfigure(fig, df):
         fig.add_trace(list_scatter[index])
     return fig
 
+def add_ROI(fig, df, k=4):
+    ''' Choose k-layers window based on energy deposited in each layer '''
+    layer_sums = df.groupby(['layer']).mipPt.sum() 
+    initial_layer = (layer_sums.rolling(window=k).sum().shift(-k+1)).idxmax()
```
bfonta, line 95: A comment would help; in particular, I do not understand why you need ```shift()```.

bye_splits/plot/display_plotly/main.py
bfonta: I mean that the figure is updated already within ```add_ROI```; returning it seems unnecessary.

bye_splits/utils/params.py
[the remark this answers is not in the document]
bfonta: Ok. @mchiusi be ready to modify this as soon as #20 is merged.

bye_splits/plot/display_plotly/event_processing.py
mchiusi: Ah, now I understand. Yes maybe I can change the variable name because I'm summing up energies belonging to the same (u,v) module in multiple k-layer.

bye_splits/plot/display_plotly/main.py
mchiusi: Yes, you are right! I extend this small change to the _update_figure_ functions in the same file.

bye_splits/plot/display_plotly/event_processing.py
mchiusi: Interesting. Solved now.

isehle (approved):
Again this mostly involves geometry and your own plotting, so I only quickly reviewed most changes other than those that directly affect me. The minor changes that you've made to the run chain shouldn't affect me too much either, so this looks good to go on my side. Thanks for the PR and I'm looking forward to seeing the new plots!

Issue #34, opened 2023-03-28
Deprecated Cfg File


isehle:
The YAML configuration file referenced here has been replaced by this [single file](https://github.com/bfonta/bye_splits/blob/main/config.yaml) since #26 was merged. Currently rewriting this for the TC skimming, so I'll include the patch in the PR. 

https://github.com/bfonta/bye_splits/blob/15dcbdb0955969550f301d90d06b5a741c4af0e1/bye_splits/production/src/skim.cc#L34

isehle:
This was fixed by [42](https://github.com/bfonta/bye_splits/pull/42) along with [46](https://github.com/bfonta/bye_splits/pull/46).

Issue #32, opened 2023-03-17
Hardcoded paths in produce.cc


jbsauvan:
Hardcoded input directory in `produce.cc`
https://github.com/bfonta/bye_splits/blob/ff9e6ec554c89ef70038cc4d48460d960bf3f76e/bye_splits/production/produce.cc#L78
While we should expect it to be read from the configuration file
https://github.com/bfonta/bye_splits/blob/ff9e6ec554c89ef70038cc4d48460d960bf3f76e/bye_splits/production/prod_params.yaml#L4-L6

In addition,  at the moment it can only use the same base directory for input and output files:
https://github.com/bfonta/bye_splits/blob/ff9e6ec554c89ef70038cc4d48460d960bf3f76e/bye_splits/production/produce.cc#L92

But the input directory could be read-only (for instance if reading data in another user directory), so it would be better to have a separate base directory to store the output. And ideally this output base directory should be configurable as well.

isehle:
I've rewritten [```produce.cc```](https://github.com/isehle/bye_splits/blob/9370c3a051e8fe76ca9334b5c73a3d2fa116d6fb/bye_splits/production/produce.cc#L144) so that it can be run without any arguments, using the input information in ```prod_params.yaml```. It now accepts an optional ```--inpath``` argument if you wish to run it on a different input file from the one set in the YAML file (very useful for job submission), and the ```--particles``` argument was removed in favor of inferring it from the input file path (which requires the particle name to be somewhere in the path). The ```prod_params.yaml``` file will be deprecated by #26 so I won't open the PR yet, but this should resolve the issue.

Issue #29, opened 2023-03-07
Extreme Lag When Loading Pile Up Data


isehle:
The method used to read in the data here, notably using uproot's ```tree.arrays()``` method, is unable to handle the trigger cell data when pile up is included. Indeed even selecting just 10 events lags for over thirty minutes (and I aborted before it finished). We'll need to find another way of reading in the data.

https://github.com/bfonta/bye_splits/blob/396555886669ed3747d55897bfd0c6c3e3cbd31c/bye_splits/data_handle/event.py#L154

bfonta:
@isehle I suppose this is already fixed by #44.

PR #33, opened 2023-03-24
New chain (preliminary): ROI finder + seeding + clustering

bfonta:
This PR brings a first version of a ROI-based reconstruction chain. Much work is still required, but this aims to provide a common ground from which to develop further algorithms. It includes:
- ROI finder by @mchiusi 
- seeding on U/V space
- additional plotting for preliminary seeding studies
- clustering on the (x/z, y/z) space
- scripts for validation and final plotting

The new chain (```bye_splits/scripts/run_default_chain.py```) works in the same fashion as the now "default" chain (```bye_splits/scripts/run_default_chain.py```). 

**Minor changes**

- Rename ```data_handle.py``` to ```data_process.py``` to avoid confusion with the name of its module
- Completely remove ```Path``` to make a consistent usage of ```os.path```
- Move common selection cuts to ```bye_splits/data_handle/data_process.baseline_selection(...)```

PR #35, opened 2023-03-28
Skim tcs

isehle:
### PR Description

The primary goal was to implement a &Delta; R selection between the trigger cells and gen-level particles in the skimming step. This greatly reduces the size of the stored arrays and should make running the reconstruction chain significantly faster, solving issue #29. In addition, this PR updates the production scripts to use the YAML configuration file in place of hardcoded paths, solving issues #32 and #34. Finally, a submission script was included that allows the user to perform the skimming step as jobs launched on HTCondor. Both should be easily generalizable to submit other jobs.

### How has this been tested?

I have confirmed that the skimming test works well for both run possibilities (using the default path set in the YAML file and the ```--inpath``` command line option). They work equally as well as part of a job submission.

### Breaking changes

The formerly required argument,```--particles```, has been removed from the production step in favor of inferring the particle type from the input file path. This obviously requires the particle name to be included somewhere in the path, or the code will throw an error. Similarly, if more than one particle type is present in the path the code will also throw an error.

### Checklist


bye_splits/production/src/skim.cc
```diff
@@ -1,97 +1,196 @@
...
+template <typename T>
+std::vector<T> tcMatch(std::vector<T> tc_col, ROOT::VecOps::RVec<int> tc_matches)
+{
+  std::vector<T> matched_tcs;
+  unsigned tc_size = tc_col.size();
+  unsigned tc_match_size = tc_matches.size();
+  for (unsigned i = 0; i < tc_size; ++i)
+  {
+
+    if (i < tc_match_size && tc_matches[i] == 1)
```
bfonta, line 68: What is the point of the first condition? By construction, shouldn't the two vectors have the same size? Btw, you can replace ```&&``` for more readable ```and```.

submit_scripts/job_params.py
```diff
@@ -0,0 +1,38 @@
...
+    # File paths stored on .txt files
+    files_photons = f"{phot_submit_dir}photon_ntuples.txt"
+    files_electrons = f"{el_submit_dir}electron_ntuples.txt"
+files_pions = []
+
+
+# Number of files to process in each batch
+files_per_batch_elec = 10
+files_per_batch_phot = 10
+files_per_batch_pion = 10
```
bfonta, line 38: Please use ```.format()``` everywhere, as it is cleaner.

submit_scripts/job_params.py
```diff
@@ -0,0 +1,38 @@
+# Permissions
```
bfonta, line 1: This file is essentially a configuration file. Can you please move it to ```config.yaml```? You could move the list part (the only part you cannot conveniently do with the ```YAML``` format) to ```job_submit.py```.

submit_scripts/job_submit.py
```diff
@@ -0,0 +1,304 @@
+#!/usr/bin/env python
```
bfonta, line 1: Please move the entire ```submit_scripts``` folder to the ```production``` folder, where it logically belongs.

submit_scripts/job_submit.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+def strip_trail(batches):
+    return [[file.rstrip() for file in batch] for batch in batches]
+
+
+def setup_batches(files, files_per_batch=10, read_dir=False):
+    if not read_dir:
+        with open(files, "r") as File:
+            Lines = File.readlines()
```
bfonta, line 28: Pedantic: please stick to lower-case variables.

submit_scripts/job_submit.py
```diff
@@ -0,0 +1,304 @@
...
+            # readlines() keeps the explicit "/n" character, strip_trail removes this
+            batches = strip_trail(my_batches(Lines, files_per_batch))
+    else:
+        paths = [
+            f"{files}{file}" for file in os.listdir(files) if file.startswith("ntuple")
```
bfonta, line 33: Please use ```.format()```

submit_scripts/job_submit.py
```diff
@@ -0,0 +1,304 @@
...
+
+# Accepts a template for a full path to a file and increments the version
+def increment_version(file_path):
+    dir, file = os.path.split(file_path)
+    base, ext = os.path.splitext(file)
+    i = 0
+    file_path = f"{dir}/{base}_v{i}{ext}"
+    while os.path.exists(file_path):
+        i += 1
+        file_path = f"{dir}/{base}_v{i}{ext}"
```
bfonta, line 48: Please use ```.format()```

submit_scripts/job_submit.py
```diff
@@ -0,0 +1,304 @@
...
+    eos_home = f"/eos/user/{user[0]}/{user}/"
+    if not os.path.exists(sub_dir):
+        os.makedirs(sub_dir)
+    script_basename = os.path.basename(script).replace(".sh", "").replace(".py", "")
+
+    submit_file_name_template = f"{sub_dir}{script_basename}_submit.sh"
+
+    submit_file_name = increment_version(submit_file_name_template)
+    with open(submit_file_name, "w") as submit_file:
+        print("#!/usr/bin/env bash", file=submit_file)
```
bfonta, line 93: Why not using ```submit_file.write()```? A shorter file name might also be cleaner.

submit_scripts/job_submit.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+if __name__ == "__main__":
+    parser = optparse.OptionParser()
+    parser.add_option(
+        "--cfg", type="string", dest="param_file", help="select the parameter file"
+    )
+    (opt, args) = parser.parse_args()
+    parameters = opt.param_file
+    main(parameters)
```
bfonta, line 304: Please use ```.format()``` everywhere.

submit_scripts/job_submit.py
```diff
@@ -0,0 +1,304 @@
...
+    configs = lambda dir: dir + "configs"
+    jobs = lambda dir: dir + "jobs"
+    logs = lambda dir: dir + "logs"
+
+    for dir in [phot_dir, elec_dir, pion_dir]:
+        config_dir = configs(dir)
+        job_dir = jobs(dir)
+        log_dir = logs(dir)
+        if not os.path.exists(config_dir):
+            os.makedirs(config_dir)
```
bfonta, line 195: What about writing a function to create a directory? You repeat these lines multiple times.

submit_scripts/job_submit.py
```diff
@@ -0,0 +1,304 @@
...
+        print(
+            "export SITECONFIG_PATH=$VO_CMS_SW_DIR/SITECONF/T2_FR_GRIF_LLR/GRIF-LLR/",
+            file=submit_file,
+        )
+        print("source $VO_CMS_SW_DIR/cmsset_default.sh", file=submit_file)
+        """print('export EOS_MGM_URL="root://eosuser.cern.ch/"', file=submit_file)
+        print(f"export EOS_HOME={eos_home}", file=submit_file)
+        print("export X509_USER_PROXY={}".format(proxy), file=submit_file)
+        print(
+            "export KRB5CCNAME=FILE:/home/llr/cms/ehle/.krb5c/ehle",
```
bfonta, line 109: The username is hardcoded.

submit_scripts/job_submit.py
```diff
@@ -0,0 +1,304 @@
...
+
+    write_batch_files(batches_phot, phot_dir, script)
+    write_batch_files(batches_elec, elec_dir, script)
+    write_batch_files(batches_pion, pion_dir, script)
+
+    configs = lambda dir: dir + "configs"
+    jobs = lambda dir: dir + "jobs"
+    logs = lambda dir: dir + "logs"
+
+    for dir in [phot_dir, elec_dir, pion_dir]:
```
bfonta, line 190: Pedantic: a tuple would be better

submit_scripts/job_submit.py
```diff
@@ -0,0 +1,304 @@
...
+
+    prepare_batch_submission(phot_dir, script, proxy, user)
+    prepare_multi_job_condor(phot_dir, script, queue, proxy)
+
+    prepare_batch_submission(elec_dir, script, proxy, user)
+    prepare_multi_job_condor(elec_dir, script, queue, proxy)
+
+    if len(files_pions) > 0:
+        prepare_batch_submission(pion_dir, script, proxy, user)
+        prepare_multi_job_condor(pion_dir, script, queue, proxy)
```
bfonta, line 209: Loop + ```**dict``` syntax?

submit_scripts/job_submit.py
```diff
@@ -0,0 +1,304 @@
...
+            "pions", pion_dir, script, user=user, queue=queue, proxy=proxy, local=local
+        )
+
+
+if __name__ == "__main__":
+    parser = optparse.OptionParser()
+    parser.add_option(
+        "--cfg", type="string", dest="param_file", help="select the parameter file"
+    )
+    (opt, args) = parser.parse_args()
```
bfonta, line 302: Pedantic: the first pair of parenthesis is not required

submit_scripts/job_submit.py
```diff
@@ -0,0 +1,304 @@
...
+    parameters = opt.param_file
```
bfonta, line 303: If the parameters were written in ```config.yaml```, passing files around would not be required.

submit_scripts/skim_pu_multicluster.sh
```diff
@@ -0,0 +1,11 @@
...
+
+work_dir="/home/llr/cms/ehle/NewRepos/bye_splits/"
+cd $work_dir
+
+batch_file=$1
+while read -r line; do
+  ./produce.exe --inpath "$line"
+done <$batch_file
+
+exit 0
```
bfonta, line 11: This usage does not seem to be described in the ```README.md```. Can you please provide more details? What steps do I need to follow to launch parallel jobs for 200PU photon samples, for instance?

submit_scripts/job_submit.py
```diff
@@ -0,0 +1,304 @@
...
+    queue = parameters.queue
+
+    read_dir = parameters.read_dir
+
+    files_electrons = parameters.files_electrons
+    files_photons = parameters.files_photons
+    files_pions = parameters.files_pions
+    files_per_batch_elec = parameters.files_per_batch_elec
+    files_per_batch_phot = parameters.files_per_batch_phot
+    files_per_batch_pion = parameters.files_per_batch_pion
```
bfonta, line 268: This is not required if using parameters in ```config.yaml```.

bfonta (commented):
Thanks for the update; it will hopefully speed things up for production. 
Please link your branch to issues #32 and #34 (right tab in the issue page) so that they are closed automatically when the PR gets merged.
In general (for future PRs only), please refrain from changing the spacing/indentation everywhere; it makes code review harder.
Once the comments are addressed, I will wait for #33 and #30 to be merged before approving this one.

bye_splits/production/produce.cc
```diff
@@ -1,94 +1,191 @@
...
+    particles = "pions";
+  }
+  else
+  {
+    std::cout << "\nInpath: " << inpath << "\n";
+    std::cout << "\nSub string: " << phot_str << "\n";
+  }
+  return particles;
+}
+
```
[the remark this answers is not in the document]
isehle: This has been fixed now, it will throw an assertion error unless one and only one particle type is present in the path. As far the singular vs. plural comment, though, everything throughout the code uses the plural convention, but I specifically check against the singular because this if-statement will catch either ```<particle>```  or ```<particle>s```, while checking against the plural will break if someone uses the singular. I prefer to use this more flexible approach.

bye_splits/production/produce.cc
```diff
@@ -1,94 +1,191 @@
...
-  string particles = args["particles"].as<string>();
+  // string inpath = args["inpath"].as<string>();
   int nevents = args["nevents"].as<int>();
 
-  std::string infile = particles + "_0PU_bc_stc_hadd.root";
+  string particles = get_particles(inpath);
+  out_dir += particles + "/new_skims/";
+
+  // Grab the file basename from the /full/path
+  std::string infile = inpath.substr(inpath.find_last_of("/\\") + 1);
```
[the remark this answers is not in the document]
isehle: Yes, it's correct, I agree that it appears to be specific for Windows but it returns the correct file name on Linux as well.

bye_splits/production/produce.cc
```diff
@@ -1,94 +1,191 @@
...
   std::string events_str = nevents > 0 ? std::to_string(nevents) + "events_" : "";
-  std::string outfile = "skim_" + events_str + infile;
-  skim(tree_name, dir + infile, dir + outfile, particles, nevents);
+  std::string outfile = outfile_base_name + "_" + events_str + infile;
+  std::string outpath = out_dir + outfile;
+
+  ifstream file;
+  file.open(outpath);
```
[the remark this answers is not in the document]
isehle: [This](https://stackoverflow.com/questions/748014/do-i-need-to-manually-close-an-ifstream) suggests otherwise, but I'm happy to do it if you prefer.

bye_splits/production/src/skim.cc
```diff
@@ -1,97 +1,196 @@
...
-	dphi = fabs(clphi[j] - gphi[0]);
-	if(dphi > M_PI) dphi -= (2 * M_PI);
-	deltaR[j] = sqrtf(dphi*dphi + deta*deta);
+
+  for (unsigned j = 0; j < ncl; ++j)
+  {
+    deta = fabs(cleta[j] - geta[0]);
+    dphi = fabs(clphi[j] - gphi[0]);
+    if (dphi > M_PI)
+      dphi -= (2 * M_PI);
```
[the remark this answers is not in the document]
isehle: I agree, though that is in fact your code—it's wrapped in brackets now.

bye_splits/production/src/skim.cc
```diff
@@ -1,97 +1,196 @@
...
+
+  for (unsigned j = 0; j < ntc; ++j)
+  {
+    float tc_x_over_z = tcx[j] / tcz[j];
+    float tc_y_over_z = tcy[j] / tcz[j];
+
+    dx = fabs(tc_x_over_z - gen_x_over_z);
+    dy = fabs(tc_y_over_z - gen_y_over_z);
+
+    deltaR[j] = sqrtf(dx * dx + dy * dy);
```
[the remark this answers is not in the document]
isehle: Updated, though I should say that I haven't noticed a difference in timing since doing so (haven't run any tests, but they feel essentially equivalent).

bye_splits/production/src/skim.cc
```diff
@@ -1,97 +1,196 @@
...
   }
 
   return deltaR;
 }
 
-ROOT::RDF::RResultPtr<long long unsigned> addProgressBar(ROOT::RDF::RNode df) {
+template <typename T>
+std::vector<T> tcMatch(std::vector<T> tc_col, ROOT::VecOps::RVec<int> tc_matches)
```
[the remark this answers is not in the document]
isehle: It's constructed as a vector of ```int```s, though it's essentially equivalent (```{0: false, 1: true}```).

bye_splits/production/src/skim.cc
isehle: Yes, actually. I've just converted the PR to Draft for this reason because I suspect this is a bigger problem than it might appear (not including it yields Segmentation Faults, and including an ```assert()``` statement on their size equivalence confirms that they aren't always the same size, breaking the code).

bye_splits/production/produce.cc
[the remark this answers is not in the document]
bfonta: If you read the comments and the other answers you'll see that the most correct approach is very situation-dependent. If you close the file you are being explicit (I'm biased by the Zen of Python!) and, importantly, you are freeing resources as soon as possible. Additionally, I see no reason not to do it.
For our use case it makes no difference (the program returns soon after), so feel free to pick your favorite approach.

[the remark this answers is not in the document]
isehle: Okay no worries, I'll include it in the next commit

bye_splits/production/src/skim.cc
isehle: This has been solved by the latest commit. The function is no longer used, opting instead to create the ```tc_deltaR``` branch and then simply add the ```tc_deltaR <= <thresh>``` condition to the [```cond_tc```](https://github.com/bfonta/bye_splits/blob/15dcbdb0955969550f301d90d06b5a741c4af0e1/bye_splits/production/src/skim.cc#L85) selection.

submit_scripts/job_submit.py
isehle: About to push a commit where this is done in a loop, yup.

submit_scripts/skim_pu_multicluster.sh
isehle: Apologies, yes I neglected to update the README for the HTCondor related part. Done now, coming in the next commit.

submit_scripts/job_submit.py
isehle: I've switched to ```.write()``` (coming in the next commit), but the filename is quite deliberate—my reasoning is explained in the ```README``` (also coming in the next commit).

isehle:
To help with another feature (running the cluster size study on HTCondor), I rewrote [```job_submit.py```](https://github.com/isehle/bye_splits/blob/skim_tcs/bye_splits/production/submit_scripts/job_submit.py) in terms of classes, as well as added a few general updates. You can see the updated version [here](https://github.com/isehle/bye_splits/blob/cl_size/pile_up/bye_splits/production/submit_scripts/job_submit.py). This update should allow for more flexibility and removes some duplication that was done in the previous version. Let me know if you would like me to integrate this version into this PR (I've confirmed that both versions run as intended), or if you would prefer to merge as-is now, and update in an upcoming PR.

isehle:
To help with another feature (running the cluster size study on HTCondor), I rewrote [```job_submit.py```](https://github.com/isehle/bye_splits/blob/skim_tcs/bye_splits/production/submit_scripts/job_submit.py) in terms of classes, as well as added a few general updates. You can see the updated version [here](https://github.com/isehle/bye_splits/blob/cl_size/pile_up/bye_splits/production/submit_scripts/job_submit.py). This update should allow for more flexibility and removes some duplication that was done in the previous version. Let me know if you would like me to integrate this version into this PR (I've confirmed that both versions run as intended), or if you would prefer to merge as-is now, and update in an upcoming PR.

isehle:
This PR was closed because the work performed was absorbed into a [new branch](https://github.com/isehle/bye_splits/tree/layerOpt) which also contains an update to the Dash Application and the calculation of energy weights. A new PR will be opened from this branch once it's been properly cleaned up.

PR #37, opened 2023-04-09
Full ROI Chain

bfonta:
This PR brings the full ROI chain with a new seeding:
- fully functional hexagonal seeding
- clustering on the x/z,y/z space taking ROI seeds and ROI TCs (or all TCs if preferred) as input
- resolution and distribution plotting utilities, including interactive visualization of hexagonal grids and events

- important fix for the "max_energy" clustering algorithm (see ```bye_splits/tasks/cluster.py```): the old version assigns all TCs (withing a maximum distance) to the same, most-energetic seed. The new version correctly prioritizes seeds based on energy, so that the seeds that are not the most energetic are still assigned some TCs.

The chain was tested with multiple parameters from start to finish.

Screenshot from 2023-04-09 23-24-26

Screenshot from 2023-04-09 23-26-38

### DIFF
--- a/bye_splits/tasks/roi.py
+++ b/bye_splits/tasks/roi.py
@@ -36,7 +36,6 @@ def get_roi_cylinder(ev_tc, roi_tcs):
     return ev_tc[(ev_tc.tc_wu.isin(uniqueu)) & (ev_tc.tc_wv.isin(uniquev)) &
                  (ev_tc.tc_layer <= ncee)]
 
-    
 def roi(pars, df_gen, df_cl, df_tc, **kw):
     """Waiting for ROI future algorithms..."""
     pass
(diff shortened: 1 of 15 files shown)


config.yaml
```diff
@@ -48,7 +48,7 @@ varEvents:
     z: 'good_tc_z'
     phi: 'good_tc_phi'
     eta: 'good_tc_eta'
-    idcl: 'good_tc_multicluster_id'
+    idcl: 'good_tc_cluster_id'
```
isehle, line 51: Was this change done on purpose? We want the `multicluster_id`, yes?

config.yaml
```diff
@@ -159,13 +161,19 @@ valid:
   FillOutComp: 'fill_comp'
   FillOut: 'fill'
 
-valid_roi:
+valid_cluster:
   ClusterOutForValidation: 'cluster_valid_roi'
```
isehle, line 165: Shouldn't we have one cluster validation that doesn't include the `_roi` tag, for the default chain?

isehle (commented):
Overall it's looking really great! Passed through the ROI related files quite quickly, but the updates elsewhere look good and shouldn't affect anything on my side. One general question—you switched from the ```with <file> as <var>``` syntax to ```<var> = <file>``` in a few places; is this primarily for debugging purposes or is there a deeper reason?

config.yaml
bfonta: You are right, I keep modifying it for local tests.

bfonta:
@isehle No important reason; I did it simply to avoid the hard-to-read indentation when one has multiple ```open()``` calls in the same ```with``` clause. By the way, we now have to explicitly call ```.close()```.

config.yaml
bfonta: [We do](https://github.com/bfonta/bye_splits/blob/feature/seeding/config.yaml#L159-L169). I changed the naming just to clarify to which part of the reco chain this validation corresponds; ```valid_roi``` was too generic, and could refer to the ROI finder, seeding or clustering in the ROI chain.

isehle: Ah yes, I missed the first one. For consistencies sake though, I'd suggest they both use the same name (either `ClusterOutValidation` or `ClusterOutForValidation`, but not both—unless there's a specific reason why not.)

isehle:
@bfonta I see, yes that's helpful enough. And yup, I'm aware that `.close()` is necessary now, but thanks for the reminder!

config.yaml
bfonta: Fair enough.

isehle (approved):
Looks great!

isehle (approved):

bye_splits/tasks/roi.py
```diff
@@ -52,21 +51,21 @@ def roi_dummy_calculator(tcs, k=4, threshold=20, nearby=True):
...
+    roi_df = pd.DataFrame()
+    
     module_sums = create_module_sums(input_df)
     module_ROI = list(module_sums[module_sums.values >= threshold].index)
-
     if nearby:
-        selected_modules = []
-        for module in module_ROI:
-            nearby_modules = [(module[0]+i, module[1]+j) for i in [-1, 0, 1] for j in [-1, 0, 1] if i*j >= 0]
-            skim = (module_sums.index.isin(nearby_modules)) & (module_sums.values > 0.3 * module_sums[module]) & (module_sums.values > 10)
```
mchiusi, line 63: Have you decided to pass every time all the six nearby modules _plus_ the central one?
Do you apply an energy selection?

mchiusi (commented):
Overall it's looking great.

bye_splits/tasks/roi.py
mchiusi: Ok perfect. That's all I wanted to know.

mchiusi (approved):

PR #42, opened 2023-05-24
Skimming: C++ to Python Translation

bfonta:
Translation of the current ```C++``` skimming step to ```python```. The code should be equivalent, namely still using ```RDataFrame```, now via ```PyROOT```. 

Advantages:
- convenience: the code can be easily integrated with the other parts of the framework, and no compilation and linking steps are required (no more issues with external dependencies like ```yaml``` and ```boost```!); in addition there is the option of using ```RDataFrame``` with ```numpy``` and ```numba```, as described [here](https://root.cern.ch/doc/master/classROOT_1_1RDataFrame.html)
- no expected reduction in performance: ```RDataFrame``` calls are still expressed in ```C++```; functions are compiled on-the-fly with ```ROOT.gInterpreter.Declare(...)```.
- future developments: this translation enables future usage of distributed technologies that can be now connected to ```RDataFrame```: ```Spark``` and ```Dask```. We should understand how they compare to ```HTCondor```, since they seem much easier to implement (see [this post](https://ep-news.web.cern.ch/content/distributed-analysis-root-rdataframe)) @isehle 

The skimming was successfully tested without multi-threading for 1k events and with multi-threading for 30k events. Warning: skimming many events with one single core is slow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,9 @@
 # Table of Contents
 
 1.  [Installation](#org0b1e512)
-2.  [Data production](#orga49b706)
-    1.  [Skimming: install `yaml-cpp` dependency](#org18f77dd)
-    2.  [Skimming: run](#org5746685)
-    3.  [Data sources](#orgfc86ff0)
+2.  [Data production](#dataprod)
+    1.  [Skimming](#skim)
+    2.  [Data sources](#sources)
 3.  [Reconstruction Chain](#org0bc224d)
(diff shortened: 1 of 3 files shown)


mchiusi (approved):
This looks great! Definitely more readable than the previous version we had!

isehle (approved):
This will be great, thanks so much for doing it!

PR #43, opened 2023-05-29
Introduce general statistics and plotting classes + Addressing comments on the u/v seeding

bfonta:
This PR addresses all comments made to my first presentation on the new seeding on (U,V) coordinates, namely:
+ [X] use Clopper-Pearson intervals for efficiency errors instead of Poisson
+ [X] add TCs to efficiency plots but cutting the noise
    + I've applied a cylinder around the center of the shower
+ [X] visualize resolution distributions also with violin plots
+ [X] calculate the seeding efficiency having as reference the trigger cells around the generated particle, skipping the ROI step. This allows to factor the ROI out of seeding efficiency measurements.

Additionally, the PR includes a new ```Collector``` and ```ChainPlotter``` classes which will hopefully represent the first step towards unifying the efficiency/resolution statistics collection and plots which are common to all our efforts. If possible, I would very much like most plotting being done through these classes, which would slowly evolve depending on the need of everybody. For instance, I would be happy to add other libraries (besides ```bokeh```). The rationale is to avoid duplicating code when we all do similar plots, such as efficiencies, binned distributions, ...

A new "meta" chain script was added to run individual chains and plots results of multiple chains together, for instance:

Screenshot from 2023-06-01 18-21-33

Finally, I've added the number of events as an additional command line argument; this ensures all intermediate files stored are different for different number of input events, which is both convenient and safe (no confusion can arise in case one decides to run only part of the chain for a specific study).

As an example, the meta chain can be run as follows:

```
python bye_splits/scripts/run_meta.py --nevents 4000
```

Many flags are there for fine-grained running. As an example, to do the data collection and plotting steps a second time without reprocessing the same data:

```
python bye_splits/scripts/run_meta.py --nevents 4000 --no_roi --no_seed --no_cluster --no_smooth --no_fill
```


bye_splits/tasks/validation.py
```diff
@@ -58,175 +60,359 @@ def validation(pars, **kw):
...
+            outgen.update( {'def': common.fill_path(self.cfg['fill']['FillOutGenCl'], **pars)})
+        if mroi:
+            extra = common.seed_extra_name(self.cfg)
+            outseed.update({'roi': common.fill_path(self.cfg['seed_roi']['SeedOut'] + extra,            **pars)})
+            outtc.update(  {'roi': common.fill_path(self.cfg['roi'][self.cfg['seed_roi']['InputName']], **pars)})
+            outgen.update( {'roi': common.fill_path(self.cfg['roi']['ROIclOut'],                        **pars)})
+            
+        sseed = {k: h5py.File(  v, mode='r') for k,v in outseed.items()}
+        stc   = {k: pd.HDFStore(v, mode='r') for k,v in outtc.items()}
+        sgen  = {k: pd.HDFStore(v, mode='r') for k,v in outgen.items()}
```
mchiusi, line 145: Is it necessary to use different libraries (`h5py` and `pandas`) to open the `HDF5` binary data format?
See next comment

bye_splits/tasks/validation.py
```diff
@@ -58,175 +60,359 @@ def validation(pars, **kw):
...
+        df['nclusters'] = self.nclusters
+        return df
+
+class Collector:
+    def __init__(self):
+        with open(params.CfgPath, 'r') as afile:
+            self.cfg = yaml.safe_load(afile)
+
+    def collect_seed(self, pars, chain_mode, debug=False, **kw):
+        """Statistics collector for ROI seeding results."""
```
mchiusi, line 127: Please explain better which is the structure of this function (_at least_ in the PR). It is quite nested and it took me some time to understand the _dictionary of dictionary_

bye_splits/tasks/validation.py
```diff
@@ -58,175 +60,359 @@ def validation(pars, **kw):
...
+
+            assert len(kseeds)==len(ktc)
+            search_ev = '{}_([0-9]{{1,7}})_ev'.format(kw['FesAlgo'])
+            for ks,kt in zip(kseeds,ktc):
+                evn = re.search(search_ev, ks).group(1)
+                assert evn in kt
+                    
+                dftc = stc[chain][kt]
+                dfseed = sseed[chain][ks]
+                colsseed = list(dfseed.attrs['columns'])
```
mchiusi, line 164: - Can you explain to me what are `ktc`, `ks`, `kt`, etc. 
 - How is `sseed[chain]` structured, i.e. how is the information packed in the `.hdf` files?

mchiusi (changes requested):

bye_splits/data_handle/data_process.py
```diff
@@ -88,7 +92,7 @@ def get_data_reco_chain_start(nevents=500, reprocess=False, tag='chain'):
         "good_tc_z": "tc_z",
         "good_tc_eta": "tc_eta",
         "good_tc_phi": "tc_phi",
-        "good_tc_cluster_id": "tc_cluster_id",
+        "good_tc_multicluster_id": "tc_multicluster_id",
```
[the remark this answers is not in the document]
bfonta: Although currently we are not using it (besides on data processing steps), this id allows to associate trigger cells to clusters, and can thus be used for (at least) two things: 
 - Validation. Imagine a scenario where we develop a new clustering algorithm which performs differently than the current one. The most fine-grained why to study the differences would be to check, for each trigger cell, which cluster was it associated to, and compare it with our new custom method.
 - Event displays. The only way to display TCs association to "external" (CMSSW) clusters is to use this id, because the clustering was not done by us, and so we do not know the association *a priori*.

bye_splits/tasks/validation.py
bfonta: This is a consequence of how those datasets were *written*. 
The question then becomes: why did I use different libraries to write the files? Ideally we would use only one, but ```pandas``` is easier to use and ```h5py``` is faster (since I can use ```numpy``` arrays directly). For instance, doing a ```groupby``` without pandas is tricky, but at the same time I want to avoid as much as possible using ```pandas``` dataframes inside the event loops of the reconstruction chain.
I would keep it this way, unless some volunteer wants to adapt everything to using ```hdf5``` only.

bye_splits/tasks/validation.py
bfonta: In order:
- I thought the code was self-explanatory here; let me know how I can improve it otherwise. ```kseeds``` and ```ktc``` are the *keys* of, respectively, the seed and TC datasets (```keys()``` method). ```ks```, ```kt```, etc, are simply the iterators of those keys, *i.e.*, they point to a single key as should be clear from the ```for``` loop. Each key represents an event, hence the loop.
- ```chain``` refers to the type of chain you are running. Currently we have 2: the "standard", which I call "def" from "default", and the new u/v, coarse seeding one (called "roi"). With this structure you can plot quantities coming from different chains, which is very useful for comparisons, and was used for my u/v seeding slides. The structure of the hdf5 files did not change, I am now simply using a dictionary of those structures, where each element corresponds a different chain. See how I efined ```outseed```, for instance.

isehle (commented):
Thanks for all the work you put into this! Besides my one question in `cluster.py`, this should all be fine for me. Would you also mind putting an example of how we can use the new plotting class in the PR? I.e., a command to run and its output, for example.

bye_splits/tasks/cluster.py
```diff
@@ -17,21 +17,21 @@
...
-    assert len(seed_keys) == len(tc_keys)
+    skeys  = [x for x in sseeds.keys() if '_ev' in x and 'central' not in x]
+    tckeys = [x for x in stc.keys()    if '_tc' in x and 'central' not in x]
 
+    assert len(skeys) == len(tckeys)
     radiusCoeffB = kw["CoeffB"]
     empty_seeds = 0
 
-    for tck, seedk in zip(tc_keys, seed_keys):
+    for tck, seedk in tqdm(zip(tckeys, skeys), total=len(tckeys)):
```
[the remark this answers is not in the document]
bfonta: ```tqdm``` is just a Python package. It will thus not interfere at all with HTCondor; you just have to make sure you have it in the installation HTCondor has access to (which is likely the one available in the submission machine).
If you are worried, do a quick test on HTCondor with some prints within a ```tqdm``` loop.

bfonta:
@isehle See ```scripts/run_meta.py```, which should exemplify how to use the two new classes.
Please note that I am aware there are bits in the new classes that are not general, and some adaptations will likely be needed as more people use them and as we understand our needs. If there is something you do not like or think could be improved, now is the moment!

bye_splits/tasks/validation.py
bfonta: I see this is related to one of the other comments. I personally think there is no much sense in documenting code in a PR, since no one will be looking at the PR when reading code (just finding the PR takes time, and people are busy).
The paremeter ```chain_mode``` is clearly understood by looking at ```self._decode_modes(self, mode)``` (I've added a doc string there). ```pars``` means always the same in the repo, but I agree the name could be better, so I've changed it to ```cli_pars```, as a reference to the command line. I've also replace the 'roi' chain by the 'cs' ("coarse seeding") chain, since that's how it is called now.

isehle (approved):

mchiusi (approved):
This really looks great! I did fetch your branch to clarify some critical points - especially in the validation - by running the code in local. 
The display code in written using `bokeh` seems _at least to me_ much more convoluted that the same code in other libraries. Anyway I tested and it works.

bfonta:
@mchiusi I am not enforcing any library, you can use the one you prefer. I would just like that all the code right before plotting is the same for all of us. We can easily add some bokeh/plotly choice later on, as we are already doing in the geometry output.

bfonta:
I am keeping the three commits for clarity, since they are completely independent.

PR #44, opened 2023-06-19
Enhancements and modifications in the display app

mchiusi:
Major changes in the webApp (deployed [here](https://display-hgcal-event-mchiusi-dashapp.app.cern.ch)):

- display code:
   - **scintillator tiles** have been added in the 3D app. The trigger cells (TCs) geometry of the entire calorimeter is now complete: **pions** and **taus** can be studied.
   - a slider has been added in the display app in order to control live the **clustering radius**. Currently the app is using the reconstruction chain implemented in CMSSW (_default reconstruction chain_ in the code) which in our framework has been translated in Python. In the future the new clustering algorithm, which includes the new _coarse seeding_ and the _seeding_ algorithms, can be easily implemented.
   - the new switch button _layer selection_ allows to visualise specific portions of the calorimeter
   - the new **coarse seeding** algorithm can be run through the application and the coarse seeds visualised on screen. More details about it can be found [here](https://indico.cern.ch/event/1279953/#3-seeding-ideas-coarse-seeding).
   - in addition to the energy scale set by default in the display app, it is possible to visualise with discrete colours the different seeds found by the reconstruction chain event by event. This functionality becomes particularly interesting when 200 pile-up photon or tau samples are used.
   - the web page layout has been updated

- `geometry.py`:
   - a new _plotly_ friendly format for the geometry can be chosen for the geometry dataframe, as alternative to _bokeh_.
   - added the function `_display_sci_cartesian` to calculate, as it is currently done for silicon, the position of the scintillator tiles in cartesian coordinates. This allows to use the same function in the display code to plot both silicon cells and tiles.
   - minor changes

- `cluster.py`:
   - the clustering function has been improved, adding a second output: a collection - in natural coordinates - of the clustered trigger cells.

- data handling:
   - in `data_process.py`, added a new method to save events after processing. Currently, to speed up the event processing through the reconstruction chain, each event from `.parquet` file is pre-processed and saved in a `.hdf` file. When a _random event_ is requested from the display app the `.hdf` file is read. The display process lasts less than 1 second. If a particular event requested by the user is not yet pre-processed, the code finds the event in the parquet file and processes it live. The entire process lasts ~3 seconds.
   - The pre-processed events are saved in the `.hdf` file using their event number as _primary key_ and the clustering radius as _secondary key_, i.e. the HDF keys are in the form of <event_number> / <cluster_radius> .
   
- production:
   - added new `genparticle_pt` branch in the skimming file (`produce.py`).

The `README.md` has been updated especially in the section _3D display app_

The new look of the web application. A 0 pile-up pion is shown. Scintillator tiles are clearly visible.
image
200 pile-up photon: a large cut (in cartesian coordinates) around the generated particle has been applied. Multiple seeds in different colours shown.
image
The same 200 pile-up photon analysed with different clustering radii.
image 
image


bfonta:
@mchiusi Thank you for the PR. Nice to see that the event displays we've been using for quite some time already will be part of ```master```. Some general comments (more detailed ones will come after merging #43 to make sure both PRs are consistent):

- Could you clean a bit your git history?
- The slides that controls the clustering radius changes one parameter, while its visualization gives the idea that it is an interval. Could you change the way you control the clustering radius so that it is clear we are changing a single number? On the other hand, this slider type works very well for the layer selection.
- I suppose you could update the "ROI" label to "CS" or similar?
- The ROIs can be selected also for the pion dataset, but nothing happens. Could you add a check when no ROIs are available?
- I see no modifications in ```tasks/roi.py```. Are you using the same version as in the ```master``` branch?
- You changed ```tasks/cluster.py```: 1) It is tricky, as you have to make sure it does not break existing code, which will happen since you've changed what the function returns 2) The "new" information you are returning is available in the output files of the task 3) I agree that it would be nice to also return outputs directly from one function to another instead of relying always on the output files, but that will have to be done for all the tasks for consistency (to allow chained calls like ```cluster(seed(roi()))```), not just for one
- As a suggestion, you could add a ```quick_random()``` function that picks an event randomly from the pool of events that were already processed (in the ```.hdf5``` file) when the user is impatient
- Please clarify in the PR what you mean by "primary" and "secondary" keys

bye_splits/tasks/fill.py
```diff
@@ -103,7 +103,8 @@ def fill(pars, df_gen, df_cl, df_tc, **kw):
             store[store_str].attrs['doc'] = 'R/z vs. Phi histo Info'
 
             cols_keep = ['Rz_bin',  'tc_phi_bin', 'tc_layer', 'tc_mipPt', 'tc_pt',
-                         'tc_x', 'tc_y', 'tc_z', 'tc_eta', 'tc_phi']
+                         'tc_wu', 'tc_wv', 'tc_cu', 'tc_cv',
```
bfonta, line 106: Why weren't these needed before? What changed?

bye_splits/tasks/seed.py
```diff
@@ -10,6 +10,7 @@
 import bye_splits
 from bye_splits.utils import common, params
 
+import yaml
```
bfonta, line 13: remove

bye_splits/utils/params.py
```diff
@@ -20,6 +20,6 @@ def read_task_params(task):
         d.update(cfg[task])
     return d
 
-LocalStorage = os.path.join(parent_dir, "data/new_algos")
+LocalStorage = os.path.join(parent_dir, "data/new_algos/")
```
bfonta, line 23: the slash should not be needed inside ```os.path.join()``` IIRC

config.yaml
```diff
@@ -1,13 +1,16 @@
 io:
-  dirphotons: FloatingpointMixedbcstcrealsig4DummyHistomaxxydr015GenmatchGenclustersntuple
+  dirphotons: FloatingpointMixedbcstcrealsig4DummyHistomaxxydr015GenmatchGenclustersntuple #l1tHGCalTriggerNtuplizer 
   treephotons: HGCalTriggerNtuple
-  filephotons: /eos/user/b/bfontana/FPGAs/new_algos/skim_with_pt_ntuple_1.root #skim_photons_0PU_bc_stc_hadd.root
-  direlectrons: hgcalTriggerNtuplizer
+  filephotons: /eos/user/i/iehle/data/PU0/photons/skim_photons_0PU_bc_stc_hadd.root #skim_photons_0PU_bc_stc_hadd.root
+  dirphotons_PU: FloatingpointMixedbcstcrealsig4DummyHistomaxxydr015GenmatchGenclustersntuple #hgcalTriggerNtuplizer
+  treephotons_PU: HGCalTriggerNtuple
+  filephotons_PU: /home/llr/cms/mchiusi/building_ROI/inputFiles/skim_photons_hadd.root #/eos/user/m/mchiusi/visualization/display3d/skim_photons_hadd_200PU.root
```
bfonta, line 7: Maybe a list with two entries for pile-up and no pile-up instead of adding so many independent variables?

config.yaml
```diff
@@ -1,13 +1,16 @@
...
+  direlectrons: FloatingpointMixedbcstcrealsig4DummyHistomaxxydr015GenmatchGenclustersntuple
   treeelectrons: HGCalTriggerNtuple
   fileelectrons: /eos/user/i/iehle/data/PU0/electrons/skim_electrons_0PU_bc_stc_hadd.root
   dirpions: hgcalTriggerNtuplizer
   treepions: HGCalTriggerNtuple
-  filepions: /eos/user/i/iehle/data/PU0/pions/skim_pions_0PU_new_hadd.root
+  filepions: /data_CMS/cms/ehle/L1HGCAL/PU0/pions/skims/skim_tightTC_dRxy_hadd.root #/eos/user/i/iehle/data/PU0/pions/skim_pions_0PU_new_hadd.root
```
bfonta, line 13: @isehle Are these paths still up-to-date?

config.yaml
```diff
@@ -19,14 +22,15 @@ geometry:
     nlayersCEH: 22
   
 selection:
-    particles: photons
-    disconnectedTriggerLayers: [2,4,6,8,10,12,14,16,18,20,22,24,26,28]
+    particles: photons_PU
+    disconnectedTriggerLayers: [2,4,6,8,10,12,14,16,18,20,22,24,26,28,30,32,34,36,38,40,42,44,46,48,50]
```
bfonta, line 26: I think the CE-H has TCs in all layers, not only even.

config.yaml
```diff
@@ -19,14 +22,15 @@ geometry:
...
     reachedEE: 2
-    deltarThreshold: 0.05
+    deltarThreshold: 0.1
```
bfonta, line 28: Rationale?

requirements.txt
```diff
@@ -1,8 +1,116 @@
...
+traitlets==5.8.1
+typing_extensions==4.4.0
+uproot==5.0.2
+wcwidth==0.2.6
+webencodings==0.5.1
+websocket-client==1.4.2
+Werkzeug==2.2.2
+wheel==0.38.4
+xyzservices==2022.9.0
+zipp==3.11.0
```
bfonta, line 116: These requirements are too specific. Please make them more flexible as they were before, with ``>``` instead of ```==```(for instance).

bfonta (commented):
### Second review

Regarding you PR text, please fix the following:
- Please include a reply to my previous comments in this thread for future clarity
- At least some of my comments seem not to be implemented in the web app that you link; can you please update the version shared via ```*.app.cern.ch``` (and the plots in the PR description)?
- The scintillator tiles were actually added in #26, and the geometry was already complete; what you did was to include them in the 3D visualization, which is very different from a technical point of view
- "the new coarse seeding algorithm has been implemented in the app": In reality no algorithm is implemented in the app; the app takes advantage of the existing algorithms, which are completely independent from the app. Could you please rephrase?
- "the graphic has been updated": what exactly? Same for "minor issues have been fixed", if you do not list what you did is simpler just to remove them. All your modifications can anyways always be seen in the commit diffs.
- "added the possibility to chose between bokeh and plotly" The mechanism to support more libraries [was already present](https://github.com/bfonta/bye_splits/blob/b721fd0c310f833f8cdbe01ad4dff9cb03ddfa4b/bye_splits/data_handle/geometry.py#L292-L304). You could say that you added support for plotly.
- In the first bullet point of "data handling.py", it is not clear to me from the text what you added and what was already available

bye_splits/data_handle/data_process.py
```diff
@@ -92,7 +102,7 @@ def get_data_reco_chain_start(nevents=500, reprocess=False, tag='chain'):
         "good_tc_z": "tc_z",
         "good_tc_eta": "tc_eta",
         "good_tc_phi": "tc_phi",
-        "good_tc_multicluster_id": "tc_multicluster_id",
+        #"good_tc_multicluster_id": "tc_multicluster_id",
```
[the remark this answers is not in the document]
mchiusi: From what I understand, this variable does not contain accurate information. The cluster count information is in `seed_idx` (which is added after running the reconstruction chain and it depends on the clustering radius).

bye_splits/data_handle/data_process.py
```diff
@@ -104,7 +114,7 @@ def get_data_reco_chain_start(nevents=500, reprocess=False, tag='chain'):
         "good_genpart_exeta":  "gen_eta",
         "good_genpart_exphi":  "gen_phi",
         "good_genpart_energy": "gen_en",
-        "good_genpart_pt":     "gen_pt",
+        #"good_genpart_pt":     "gen_pt",
```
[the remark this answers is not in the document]
mchiusi: `gen_pt` can be easily calculated from `gen_en` and `gen_eta`. No need to store it in the HDF file.

bye_splits/data_handle/data_process.py
```diff
@@ -121,13 +131,78 @@ def get_data_reco_chain_start(nevents=500, reprocess=False, tag='chain'):
     ds_cl = ds_cl.rename(columns=cl_keep)
     return ds_gen, ds_cl, ds_tc
 
+def colorscale(df, variable, scale, saturate=False):
```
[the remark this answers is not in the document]
mchiusi: Thanks a lot for the suggestion! I will do it

bye_splits/data_handle/data_process.py
```diff
@@ -121,13 +131,78 @@ def get_data_reco_chain_start(nevents=500, reprocess=False, tag='chain'):
...
+    norm_points = (df[variable]-df[variable].min())/(df[variable].max()-df[variable].min())
+    colorscale = sample_colorscale(scale, norm_points) 
+    if saturate:
+        ["rgb(255,255,255)" if norm_points[i] == 1 else s for i, s in enumerate(colorscale)]
+    return colorscale
+
+def get_event(filename, event, coefs):
```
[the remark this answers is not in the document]
mchiusi: Moved to `bye_splits/plot/display_plotly/processing.py`

bye_splits/data_handle/data_process.py
```diff
@@ -121,13 +131,78 @@ def get_data_reco_chain_start(nevents=500, reprocess=False, tag='chain'):
...
+    if isinstance(data, dict):
+        for key, item in data.items():
+            if isinstance(item, pd.DataFrame):
+                item.to_hdf(h5file, path + str(key))
+            else:
+                raise ValueError('Cannot save %s type'%type(item))
+    else:
+        data.to_hdf(h5file, path)
+
+def process_and_store(dict_events, gen_info, ds_geom, h5file):
```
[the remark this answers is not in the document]
mchiusi: You're right. Now the functions are organise in the following way:
 - main calls `get_data` in `processing.py` file
 - `get_data` calls in turn `process_event()` if the reco chain has to be called and re-runned or `get_event()` if nothing has to be reprocessed

bye_splits/data_handle/data_process.py
```diff
@@ -121,13 +131,78 @@ def get_data_reco_chain_start(nevents=500, reprocess=False, tag='chain'):
...
+
+            
 def EventDataParticle(tag, reprocess, logger=None, debug=False, particles=None):
     """Factory for EventData instances of different particle types"""
     with open(params.CfgPath, "r") as afile:
         cfg = yaml.safe_load(afile)
         if particles is None:
             particles = cfg["selection"]["particles"]
-        if particles not in ("photons", "electrons", "pions"):
+        if particles not in ("photons", "photons_PU", "electrons", "pions"):
```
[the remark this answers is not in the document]
mchiusi: I agree that photons_PU is not a particle. I don't know if it's worth adding a boolean variable to be passed from the display code to `event.py`. We can discuss later.

bye_splits/data_handle/geometry.py
```diff
@@ -13,6 +13,7 @@
 import awkward as ak
 import uproot as up
 import pandas as pd
+pd.options.mode.chained_assignment = None
```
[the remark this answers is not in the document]
mchiusi: Moved to the exact line causing this warning. Then I restore this variable to its default value.

bye_splits/data_handle/geometry.py
```diff
@@ -226,19 +232,20 @@ def _display_silicon_tcs(self, df):
...
         df['hex_x'] = df[xcorners_str].values.tolist()
-        df['hex_x'] = df['hex_x'].map(lambda x: [[x]])
         df['hex_y'] = df[ycorners_str].values.tolist()
-        df['hex_y'] = df['hex_y'].map(lambda x: [[x]])
+        if library == 'bokeh':
+            df['hex_x'] = df['hex_x'].map(lambda x: [[x]])
+            df['hex_y'] = df['hex_y'].map(lambda x: [[x]])
 
-        remove = ['tc_x_center', 'tc_y_center', 'wx_center', 'wy_center',
+        remove = ['tc_x_center', 'tc_y_center', #'wx_center', 'wy_center',
```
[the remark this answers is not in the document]
mchiusi: Yes indeed I'm removing the removed variables :)

bye_splits/data_handle/geometry.py
```diff
@@ -270,29 +277,52 @@ def _display_scintillator_tcs(self, df):
                (df.triggercellieta==df.min_ieta), 'rmax'] -= arc_dist/2
         
         df = df.drop(['R', 'arc', 'phi', 'min_ieta', 'max_ieta'], axis=1)
+
+        if library == 'plotly':
+            df = self._display_sci_cartesian(df)
+        df['waferu'] = -999.0
```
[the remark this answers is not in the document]
mchiusi: By default `waferu` variable for scintillator tiles is set to 0. -999 is currently used in `event.py`.
I need thier perfect combination for the [merging](https://github.com/bfonta/bye_splits/pull/44/files/0046fb3c964ba92444c6614478a61f0299d880f2#diff-a81f59cf7931c089a8364409940ed36b8bb6fa6d1be55452ffa0e971f5a573e3R187-R188).

bye_splits/plot/display_plotly/main.py
```diff
@@ -3,100 +3,131 @@
...
-    parser.add_argument('--host',type=str,default='llruicms01.in2p3.fr')
-    parser.add_argument('--port',type=int,default=8004)
+    parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('--host',type=str,default='llruicms01.in2p3.fr', help='host machine to run the application')
+    parser.add_argument('--port',type=int,default=8004, help='port to use')
+    parser.add_argument("--no_fill", action="store_true")
+    parser.add_argument("--no_smooth", action="store_true")
+    parser.add_argument("--no_seed", action="store_true")
+    parser.add_argument("--no_cluster", action="store_true")
+    parser.add_argument("--no_validation", action="store_true")
```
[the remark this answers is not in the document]
mchiusi: They are used in `scripts/run_default_chain.py` Do you want to create a function to store all these arguments and then call it in `bye_splits/plot/display_plotly//main.py` and `scripts/run_default_chain.py`

bye_splits/plot/display_plotly/main.py
```diff
@@ -3,100 +3,131 @@
...
+    parsing.add_parameters(parser)
     args = parser.parse_args()
     
-    app.run_server(debug=True,
-                   host=args.host,
-                   port=args.port)
+    process = processing.Processing() 
+    
+    app.run_server(debug=True, host=args.host, port=args.port)
```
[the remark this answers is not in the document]
mchiusi: Sorry :)

bye_splits/plot/display_plotly/plot_event.py
```diff
@@ -10,27 +10,129 @@
...
+
+def prepare_slider(df, page='3D'):
+    if page == '3D':
+        return dcc.RangeSlider(df['layer'].min(),df['layer'].max(),value=[df['layer'].min(), 
+                               df['layer'].max()], step=None, marks={int(layer) : {"label": str(int(layer))} for each,
+                               layer in enumerate(sorted(df['layer'].unique()))}, id = 'slider')
+    else:
+        return dcc.Slider(df['layer'].min(),df['layer'].max(),value=11, step=None,
+                          marks={int(layer) : {"label": str(layer)} for each,
+                          layer in enumerate(sorted(df['layer'].unique()))}, id = 'slider')
```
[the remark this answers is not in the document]
mchiusi: Thanks, you're absolutely right!

bye_splits/plot/display_plotly/plot_event.py
```diff
@@ -10,27 +10,129 @@
...
+
+def add_CS(fig, df):
+    ''' Choose k-layers window based on energy deposited in each layer '''
+    roi_df, _ = roi_dummy_calculator(df, threshold=[170, 180, 200])
```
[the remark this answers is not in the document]
mchiusi: For the moment, I've configured it this way because the idea is to choose an optimal configuration for the algorithm and continue using it. I'm unsure if it's worth implementing a slider to adjust the threshold in real-time from the GUI.

bye_splits/plot/display_plotly/plot_event.py
```diff
@@ -10,27 +10,129 @@
...
+
+    list_scatter = plot_modules(roi_df)
+    for index in range(len(list_scatter)):
+        fig.add_trace(list_scatter[index])
+
+################################################
+############## Useful conversions ##############
+
+def sph2cart(eta, phi, z=322.):
```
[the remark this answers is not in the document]
mchiusi: In 3D space, keeping constant eta and phi means finding a straight line. To calculate x and y you need the z coordinate. In this case is always the same value: the z coordinate at the beginning of the HGCAL

bye_splits/plot/display_plotly/processing.py
```diff
@@ -0,0 +1,99 @@
...
+        
+        self.reprocess = self.cfg['3Ddisplay']['reprocess']
+        self.coefs = self.cfg['3Ddisplay']['coefs']
+        self.geom_data = GeometryData(reprocess=False, logger=log)
+        self.ds_geom = self.geom_data.provide(library='plotly')
+        self.filename = None
+        self.list_events = None
+
+    def random_event(self, f):
+        return random.choice(self.list_events)
```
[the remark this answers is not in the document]
mchiusi: Mmm no. Should I?

bye_splits/tasks/cluster.py
```diff
@@ -133,13 +134,22 @@ def cluster(pars, in_seeds, in_tc, out_valid, out_plot, **kw):
         else:
             dfout = pd.concat((dfout, cl3d[cl3d_cols + ["event"]]), axis=0)
 
+        df_cluster.append(df[['seed_idx', 'tc_wu', 'tc_wv', 'tc_cu', 'tc_cv',
+                              'tc_layer']])
```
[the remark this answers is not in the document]
mchiusi: Good point. I will change the name to `df_tc`
`dfout`contains the information about the cluster, like energy, position. It does not contain TC information I need to plot the event!

bye_splits/tasks/fill.py
mchiusi: I don't know about _before_. Currently I'm using these variables for the merging procedure between clusterized TCs and non-clusterized ones.

bye_splits/scripts/cluster_size/run_cluster_size.py
```diff
@@ -128,8 +128,8 @@ def cluster_size(pars, cfg):
     parser.add_argument("--no_smooth", action="store_true")
     parser.add_argument("--no_seed", action="store_true")
     parser.add_argument("--no_cluster", action="store_true")
-    nevents_help = "Number of events for processing. Pass '-1' for all events."
-    parser.add_argument("-n", "--nevents", help=nevents_help, default=-1, type=int)
+    #nevents_help = "Number of events for processing. Pass '-1' for all events."
+    #parser.add_argument("-n", "--nevents", help=nevents_help, default=-1, type=int)
     parsing.add_parameters(parser)
 
     FLAGS = parser.parse_args()
```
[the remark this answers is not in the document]
isehle: Sorry for missing this—I've mistakenly been signed up to get notified for every comment on this thread, so I was ignoring all of them and missed the one I actually should have read.

In any case, it's currently set in the ```yaml``` configuration file, so having it as an argument is redundant anyway. We can just delete it.

config.yaml
isehle: I've been following [this structure](https://github.com/isehle/bye_splits/blob/07163927e9963a66aced536dd483d1855f294350/config.yaml#L2) for awhile and I think it's clean and works well (if you ignore the messiness of all the commented lines in my most recent pushed commit). Let me know if something like this would work for both of you.

bye_splits/plot/display_plotly/main.py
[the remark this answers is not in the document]
bfonta: It seems there is already duplication on ```scripts/run_default_chain.py```, ```scripts/run_meta.py``` and ```scripts/run_roi_chain.py```. Could you please move all the duplication to a separate function under ```utils/parsing.py```, and call it from the four scripts (three above plus yours)?

config.yaml
isehle: Indeed.

bye_splits/plot/display_plotly/processing.py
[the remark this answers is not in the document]
bfonta: It is always a good idea in case you want to reproduce your previous results. I am not sure, but I suppose the default is generated with the time of the machine (or similar), so every time you run the function it produces a result starting from an unknown seed.

bye_splits/utils/params.py
bfonta: I'm fine with the change as long as you are able to run both chains without errors. Can you please confirm? I'm asking this since you are changing a global parameter, which might affect things beyond the scope of this PR.

config.yaml
mchiusi: Ok I tried to put things more organised

config.yaml
mchiusi: You are definitely right!

config.yaml
bfonta: @mchiusi Please update.

config.yaml
mchiusi: Restored the original value

config.yaml
```diff
@@ -49,7 +53,7 @@ varEvents:
     z: 'good_tc_z'
     phi: 'good_tc_phi'
     eta: 'good_tc_eta'
-    idcl: 'good_tc_multicluster_id'
+    #idcl: 'good_tc_multicluster_id'
```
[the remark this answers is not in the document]
mchiusi: See the other comment above, we can discuss in these days

requirements.txt
mchiusi: Ah yes. Indeed these are actually the packages I have installed in my local conda environment. Sorry

bye_splits/plot/display_plotly/processing.py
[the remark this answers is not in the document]
mchiusi: This random_event function extracts one random event from a list of events. To reproduce the event previously extracted, the user can recall the same event directly using the proper field in the webApp.

config.yaml
mchiusi: Done! Thanks Isaac

bye_splits/data_handle/data_process.py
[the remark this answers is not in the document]
mchiusi: Corrected. Now added a variable `particles` which can assume the values of: 
 - photons
 - electrons
 - pions

and the variable `PU` which is PU0 or PU200.
Modifications have been made to the config.yaml file for completeness. Please review it.

bye_splits/plot/display_plotly/main.py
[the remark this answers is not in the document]
mchiusi: Done, please have a look. New code is in `utils/parsing.py`. The main function has been greatly simplified now.

bye_splits/data_handle/data_process.py
```diff
@@ -19,6 +20,7 @@
 from data_handle.geometry import GeometryData
 from data_handle.event import EventData
 from data_handle.data_input import InputData
+from plotly.express.colors import qualitative
```
bfonta, line 23: Petty comment: usually I split the imports into the "internal ones", i.e., the ones referring to our code, from "external" ones. To keep this consistent, please import ```plotly``` under ```hdf5``` and leave one blank line between "external" and "internal" imports.

bye_splits/data_handle/data_process.py
```diff
@@ -70,12 +72,18 @@ def baseline_selection(df_gen, df_cl, sel, **kw):
     print("The baseline selection has a {}% efficiency: {}/{}".format(np.round(eff,2), nout, nin))
     return data
 
-def get_data_reco_chain_start(nevents=500, reprocess=False, tag='chain'):
+def get_data_reco_chain_start(nevents=500, reprocess=False, tag='chain', particles='photons', PU='PU0', event=None):
```
bfonta, line 75: It is redundant to include the substring 'PU' in the argument ```PU```; you can convert it to an ```int``` to make it also more robust to invalid user input.
To be clear, I propose: ```def get_data_reco_chain_start(..., pu=0, ...)```. You'll have to add the substring 'PU' internally. I would avoid capital letters in argument names to keep the overall code style.

bye_splits/data_handle/data_process.py
```diff
@@ -70,12 +72,18 @@ def baseline_selection(df_gen, df_cl, sel, **kw):
...
-    data_part_opt = dict(tag=tag, reprocess=reprocess, debug=True)
-    data_particle = EventDataParticle(**data_part_opt)
-    ds_all, events = data_particle.provide_random_events(n=nevents, seed=42)
-    # ds_all = data_particle.provide_events(events=[170004, 170015, 170017, 170014])
+    data_particle = EventDataParticle(particles, PU, tag, reprocess, debug=True)
+    if event is None:
+        ds_all, events = data_particle.provide_random_events(n=nevents)
+        # ds_all = data_particle.provide_events(events=[170004, 170015, 170017, 170014])
+    else:
+        ds_all = data_particle.provide_event(event,merge=False)
```
bfonta, line 82: petty: add space between arguments

bye_splits/data_handle/data_process.py
```diff
@@ -70,12 +72,18 @@ def baseline_selection(df_gen, df_cl, sel, **kw):
...
     return data
 
-def get_data_reco_chain_start(nevents=500, reprocess=False, tag='chain'):
+def get_data_reco_chain_start(nevents=500, reprocess=False, tag='chain', particles='photons', PU='PU0', event=None):
     """Access event data."""
-    data_part_opt = dict(tag=tag, reprocess=reprocess, debug=True)
-    data_particle = EventDataParticle(**data_part_opt)
-    ds_all, events = data_particle.provide_random_events(n=nevents, seed=42)
-    # ds_all = data_particle.provide_events(events=[170004, 170015, 170017, 170014])
+    data_particle = EventDataParticle(particles, PU, tag, reprocess, debug=True)
```
bfonta, line 77: ```debug``` should be propagated from outside

bye_splits/data_handle/data_process.py
```diff
@@ -104,7 +111,6 @@ def get_data_reco_chain_start(nevents=500, reprocess=False, tag='chain'):
         "good_genpart_exeta":  "gen_eta",
         "good_genpart_exphi":  "gen_phi",
         "good_genpart_energy": "gen_en",
-        "good_genpart_pt":     "gen_pt",
```
bfonta, line 107: One of the things this PR does is to add the gen pT. Why are you removing it here?

bye_splits/data_handle/data_process.py
```diff
@@ -120,24 +126,23 @@ def get_data_reco_chain_start(nevents=500, reprocess=False, tag='chain'):
     ds_cl = ds_all["cl"]
     ds_cl = ds_cl.rename(columns=cl_keep)
     return ds_gen, ds_cl, ds_tc
-
-def EventDataParticle(tag, reprocess, logger=None, debug=False, particles=None):
+            
+def EventDataParticle(particles, PU, tag, reprocess, debug=False, logger=None):
```
bfonta, line 130: petty: ```pu``` instead of ```PU```

bye_splits/data_handle/event.py
```diff
@@ -155,10 +155,23 @@ def provide_random_events(self, n, seed=None):
...
 
     def select(self):
-        with up.open(self.indata.path, array_cache="550 MB", num_workers=8) as f:
+        with up.open(self.indata.path, array_cache='550 MB', num_workers=8) as f:
             tree = f[self.indata.tree_path]
+            total_events = tree.num_entries
             allvars = set([y for x in self.var.values() for y in x.values()])
-            data = tree.arrays(filter_name="/" + "|".join(allvars) + "/", library="ak")
+
+            threshold_size = 0.1
```
bfonta, line 163: as discussed, the name of the variable is not clear, and you can merge it with the following one

bye_splits/data_handle/event.py
```diff
@@ -155,10 +155,23 @@ def provide_random_events(self, n, seed=None):
...
+            threshold_size = 0.1 
+            threshold_size_bytes = threshold_size * 8e+9
+            data = ak.Array([])
+            for array in tree.iterate(filter_name='/' + '|'.join(allvars) + '/', step_size='20 MB', library='ak'):
+                if (data.layout.nbytes + array.layout.nbytes) <= threshold_size_bytes:
+                    data = ak.concatenate([data, array], axis=0)
+                else:
+                    break
+            
+            if len(data)/total_events < 0.1:
```
bfonta, line 172: For an empty input file the check crashes due to a division by zero. Please add some check, or a ```try ...  except``` statement.

bye_splits/data_handle/event.py
```diff
@@ -155,10 +155,23 @@ def provide_random_events(self, n, seed=None):
...
+                print('[WARNING] Function select() in event.py \nThe number of events in the Parquet file is less than 10% compared to the events in the ROOT file.')
```
bfonta, line 173: Replace 0.1 by a variable, and use it in the print. This increases flexibility if we nee to change the number in the future.

bye_splits/data_handle/geometry.py
```diff
@@ -270,8 +280,31 @@ def _display_scintillator_tcs(self, df):
                (df.triggercellieta==df.min_ieta), 'rmax'] -= arc_dist/2
         
         df = df.drop(['R', 'arc', 'phi', 'min_ieta', 'max_ieta'], axis=1)
+
+        if self.library == 'plotly':
+            df = self._display_plotly_sci_cartesian(df)
+        df['waferu'] = -999.0
```
bfonta, line 286: Why do you need this?
If you are in need of a dummy value, use ```None``` or ```np.nan```. It is also suspect that you are not doing the same for ```df['waferv']```.

bye_splits/data_handle/geometry.py
```diff
@@ -270,8 +280,31 @@ def _display_scintillator_tcs(self, df):
...
+                      y0.values, y1.values, y2.values, y3.values]).T
+        xcols, ycols = ['x1','x2','x3','x4'], ['y1','y2','y3','y4']
+        df_sci = pd.DataFrame(data=d, columns=xcols+ycols)
+        df['diamond_y']= df_sci[['x1','x2','x3','x4']].values.tolist()
+        df['diamond_x']= df_sci[['y1','y2','y3','y4']].values.tolist()
+
+        return df.drop(columns=['rmin', 'rmax', 'phimin', 'phimax'])
+   
+    def cil2cart(self, r, phi):
+        x,y = r*np.cos(np.pi/2-phi), r*np.sin(np.pi/2-phi)
```
bfonta, line 305: petty: add a space between x and y: ```x, y```

bye_splits/data_handle/geometry.py
```diff
@@ -289,18 +322,12 @@ def _parquet_to_geom(self, ds, section, region, lrange):
...
-                raise NotImplementedError()
-            df = self._display_trigger_cells(df, library)
-        else:
-            libraries = ('bokeh', )
-            if library not in libraries:
-                raise NotImplementedError()
-            #df = self._display_cells(df, library)
+        libraries = ('bokeh', 'plotly')
+        if self.library not in libraries:
+            raise NotImplementedError()
```
bfonta, line 329: add an error message

bye_splits/plot/display_plotly/main.py
```diff
@@ -3,100 +3,118 @@
...
 
-from dash import Dash, dcc, html, Input, Output, State, ctx
+from bye_splits.utils import parsing, common
+from dash import Dash, html, Input, Output, State, ctx
 from dash.exceptions import PreventUpdate
-import argparse
+import dash_bootstrap_components as dbc
+from dash_bootstrap_templates import load_figure_template
+import plot_event as plt
+import processing
```
bfonta, line 17: same petty comment as before: please split "internal" from "external" imports

bye_splits/plot/display_plotly/plot_event.py
```diff
@@ -8,40 +8,158 @@
...
+
+def geom_selection(df_dict):
+    gen_info = next(reversed(df_dict.values()))
+
+    for coef in list(df_dict.keys())[:-1]:
+        eta = gen_info['gen_eta'].values[0]
+        phi = gen_info['gen_phi'].values[0]
+        x_gen, y_gen = sph2cart(eta, phi)
+        df_dict[coef] = df_dict[coef][np.sqrt((x_gen-df_dict[coef].tc_x)**2+(y_gen-df_dict[coef].tc_y)**2)<50]
+
```
bfonta, line 22: Is 50 a known matching or clustering parameter?

bye_splits/plot/display_plotly/plot_event.py
```diff
@@ -8,40 +8,158 @@
...
+    return df_dict
+
+def prepare_slider(df, page='3D'):
+    slider_options = {
+        'min': df['tc_layer'].min(),
+        'max': df['tc_layer'].max(),
+        'step': None,
+        'marks': {int(layer): {"label": str(int(layer))} for each, layer in enumerate(sorted(df['tc_layer'].unique()))},
```
bfonta, line 30: petty: split in two lines

bye_splits/plot/display_plotly/plot_event.py
```diff
@@ -8,40 +8,158 @@
...
+def add_CS(fig, df):
+    ''' Choose k-layers window based on energy deposited in each layer '''
+    roi_df, _ = roi_dummy_calculator(df, threshold=[170, 180, 200])
+
+    list_scatter = plot_modules(roi_df)
+    for index in range(len(list_scatter)):
+        fig.add_trace(list_scatter[index])
+
+################################################
+############## Useful conversions ##############
```
bfonta, line 54: Add this kind of comments as doc-strings to each function. They can then be used by all kinds of documentation software.

bye_splits/plot/display_plotly/plot_event.py
```diff
@@ -8,40 +8,158 @@
...
+
+def sph2cart(eta, phi, z=322.):
+    theta = 2*np.arctan(np.exp(-eta))
+    r = z / np.cos(theta)
+    x = r * np.sin(theta) * np.cos(phi)
+    y = r * np.sin(theta) * np.sin(phi)
+    return x,y
```
bfonta, line 61: petty: add space between the arguments

bye_splits/plot/display_plotly/processing.py
```diff
@@ -0,0 +1,157 @@
...
+parent_dir = os.path.abspath(__file__ + 3 * '/..')
+sys.path.insert(0, parent_dir)
+
+from bye_splits.plot.display_plotly import yaml, np, pd, go, dbc
+from utils import params, common 
+from data_handle.data_process import *
+from data_handle.geometry import GeometryData
+from scripts.run_radii_chain import run_radii_chain
+from plotly.express.colors import sample_colorscale, qualitative
+from dash import dcc, html
```
bfonta, line 16: same comments about "internal" and "external" imports

bye_splits/plot/display_plotly/processing.py
```diff
@@ -0,0 +1,157 @@
...
+    def __init__(self):
+        with open(params.CfgPath, "r") as afile:
+            self.cfg = yaml.safe_load(afile)
+
+        self.data_particles = {'photons 0PU':     {'particles': 'photons',   'PU': 'PU0'},
+                               'photons 200PU':   {'particles': 'photons',   'PU': 'PU200'},
+                               'electrons 0PU':   {'particles': 'electrons', 'PU': 'PU0'},
+                               'electrons 200PU': {'particles': 'electrons', 'PU': 'PU200'},
+                               'pions 0PU':       {'particles': 'pions',     'PU': 'PU0'},
+                               'pions 200PU':     {'particles': 'pions',     'PU': 'PU200'}}
```
bfonta, line 35: You could define this in a 2D loop, where the variables can eventually be connected with other parts of the repository.

bye_splits/plot/display_plotly/processing.py
```diff
@@ -0,0 +1,157 @@
...
+        self.reprocess = self.cfg['3Ddisplay']['reprocess']
+        self.coefs = self.cfg['3Ddisplay']['coefs']
+        self.ds_geom = GeometryData(reprocess=False, logger=log, library='plotly').provide()
+        self.filename = None
+        self.list_events = None
+
+    def random_event(self, f):
+        return random.choice(self.list_events)
+
+    def filestore(self, particles, PU):
```
bfonta, line 46: ```PU``` -> ```pu``` all over this file

bye_splits/plot/display_plotly/processing.py
```diff
@@ -0,0 +1,157 @@
...
+                   'waferu'       : 'tc_wu',
+                   'waferv'       : 'tc_wv',
+                   'triggercellu' : 'tc_cu',
+                   'triggercellv' : 'tc_cv',
+                   'layer'        : 'tc_layer'}
+    
+        sci_update = {'triggercellieta' : 'tc_cu',
+                      'triggercelliphi' : 'tc_cv',
+                      'layer'           : 'tc_layer',
+                      'waferu'          : 'tc_wu'}
```
bfonta, line 105: How come you only need ```waferu``` and not ```waferv```?

bye_splits/plot/display_plotly/processing.py
```diff
@@ -0,0 +1,157 @@
...
+
+parent_dir = os.path.abspath(__file__ + 3 * '/..')
+sys.path.insert(0, parent_dir)
+
+from bye_splits.plot.display_plotly import yaml, np, pd, go, dbc
+from utils import params, common 
+from data_handle.data_process import *
+from data_handle.geometry import GeometryData
+from scripts.run_radii_chain import run_radii_chain
+from plotly.express.colors import sample_colorscale, qualitative
```
bfonta, line 15: ```qualitative``` is only used once, better to import directly what you need

bye_splits/scripts/run_default_chain.py
```diff
@@ -9,6 +9,7 @@
 parent_dir = os.path.abspath(__file__ + 2 * "/..")
 sys.path.insert(0, parent_dir)
 
+import yaml
```
bfonta, line 12: ```yaml``` is not used in this file; remove

bye_splits/scripts/run_radii_chain.py
bfonta, line 1: Isn't this from @isehle ?

bye_splits/tasks/cluster.py
```diff
@@ -133,15 +134,18 @@ def cluster(pars, in_seeds, in_tc, out_valid, out_plot, **kw):
         else:
             dfout = pd.concat((dfout, cl3d[cl3d_cols + ["event"]]), axis=0)
 
+        df_tc.append(df[['seed_idx', 'tc_wu', 'tc_wv', 'tc_cu', 'tc_cv','tc_layer']])
```
bfonta, line 137: petty: missing space before ```tc_layer```

bye_splits/tasks/roi.py
```diff
@@ -50,7 +50,34 @@ def roi(pars, df_gen, df_cl, df_tc, **kw):
...
+#        for im, module in enumerate(module_ROI):
+#            nearby_modules = [(module[0]+s, module[1]-r)
+#                              for s in [-1,0,1] for r in [-1,0,1] for q in [-1,0,1]
+#                              if s + r + q == 0]
+#            tc_roi = input_df[input_df.set_index(['tc_wu','tc_wv']).index.isin(nearby_modules)]
+#            with common.SupressSettingWithCopyWarning():
+#                tc_roi['roi_id'] = im
+#            roi_df = pd.concat((roi_df, tc_roi), axis=0)                
+#
+#    return roi_df, pd.DataFrame(module_ROI)
```
bfonta, line 78: Maybe remove this huge commented block?

bye_splits/tasks/roi.py
```diff
@@ -50,7 +50,34 @@ def roi(pars, df_gen, df_cl, df_tc, **kw):
...
+
+def roi_dummy_calculator(tcs, k=4, threshold=[170, 180, 200], eta_list=[1.5, 1.9, 2.3, 3.2], nearby=False):
```
bfonta, line 80: petty: split arguments tin two lines

config.yaml
```diff
@@ -19,24 +36,27 @@ geometry:
     nlayersCEH: 22
   
 selection:
-    particles: photons
+    particles: 'photons'
```
bfonta, line 39: you don't need the quotes; remove

config.yaml
```diff
@@ -19,24 +36,27 @@ geometry:
...
+      photons: [62878, 55896]
 
 varEvents:
   gen:
     event: 'event'
     geta: 'good_genpart_exeta'
     gphi: 'good_genpart_exphi'
     gen: 'good_genpart_energy'
-    gpt: 'good_genpart_pt'
+    #gpt: 'good_genpart_pt'
```
bfonta, line 59: Again, don't you have this variable? Maybe you realized you can calculate it directly from others with ```common.get_pt()```?

bfonta (changes requested):
# Third review

These are almost all very minor comments, and we can merge it as soon as you address them. Could you just provide an answer to previous review comments?

mchiusi:
> @mchiusi Thank you for the PR. Nice to see that the event displays we've been using for quite some time already will be part of `master`. Some general comments (more detailed ones will come after merging #43 to make sure both PRs are consistent):
> 
> * Could you clean a bit your git history?

I have done it uploading the second set of comments. At the moment commits are organised in macro-arguments.
> * The slider that controls the clustering radius changes one parameter, while its visualization gives the idea that it changes an interval. Could you modify the way you control the clustering radius so that it is clear we are changing a single number? On the other hand, this slider type works very well for the layer selection.

Thanks for the suggestion, now the slider has been set so that continuous values are excluded (`included = False`)
> * I suppose you could update the "ROI" label to "CS" or similar?

In the display code I remove every "ROI" label. Should I also change the name to `task/roi.py`?
> * The ROIs can be selected also for the pion dataset, but nothing happens. Could you add a check when no ROIs are available?

Now when pion events are displayed the CS switch is removed.
> * I see no modifications in `tasks/roi.py`. Are you using the same version as in the `master` branch?

I was calculating the ROI in the display code. Now I moved everything in `tasks/roi.py` and I have updated the code.
> * You changed `tasks/cluster.py`: 1) It is tricky, as you have to make sure it does not break existing code, which will happen since you've changed what the function returns 2) The "new" information you are returning is available in the output files of the task 3) I agree that it would be nice to also return outputs directly from one function to another instead of relying always on the output files, but that will have to be done for all the tasks for consistency (to allow chained calls like `cluster(seed(roi()))`), not just for one

`tasks/cluster.py` now returns the same output as before, since the additional information I have added is saved in files by the same function.
> * As a suggestion, you could add a `quick_random()` function that picks an event randomly from the pool of events that were already processed (in the `.hdf5` file) when the user is impatient

If the `.hdf5` file is present in the local folder, automatically the coed picks an event randomly from it. Why should I add another function to provide an event?
> * Please clarify in the PR what you mean by "primary" and "secondary" keys

Now it should be clearer.

mchiusi:
> ### Second review
> Regarding you PR text, please fix the following:
> 
> * Please include a reply to my previous comments in this thread for future clarity
> * At least some of my comments seem not to be implemented in the web app that you link; can you please update the version shared via `*.app.cern.ch` (and the plots in the PR description)?

Yes, now it has been implemented. The code has been slightly modified so that the input files are all stored in my `eos` repository. It has in the `webApp` branch [here](https://github.com/mchiusi/bye_splits/tree/webApp)
> * The scintillator tiles were actually added in [Scintillator geometry #26](https://github.com/bfonta/bye_splits/pull/26), and the geometry was already complete; what you did was to include them in the 3D visualization, which is very different from a technical point of view

I think it is clear from the PR what I did.
> * "the new coarse seeding algorithm has been implemented in the app": In reality no algorithm is implemented in the app; the app takes advantage of the existing algorithms, which are completely independent from the app. Could you please rephrase?

Answer in the previous set of comments. Now the code has been moved to `tasks/roi.py` and updated.
> * "the graphic has been updated": what exactly? Same for "minor issues have been fixed", if you do not list what you did is simpler just to remove them. All your modifications can anyways always be seen in the commit diffs.

Removed.
> * "added the possibility to chose between bokeh and plotly" The mechanism to support more libraries [was already present](https://github.com/bfonta/bye_splits/blob/b721fd0c310f833f8cdbe01ad4dff9cb03ddfa4b/bye_splits/data_handle/geometry.py#L292-L304). You could say that you added support for plotly.

Thanks.
> * In the first bullet point of "data handling.py", it is not clear to me from the text what you added and what was already available

Previously, there was no capability to save events after processing. I have now added this feature along with the mechanism for handling post-processed data. The existing code for searching events in the parquet file has been retained. Please let me know if I have to update the PR, it seems clear to me.

bye_splits/data_handle/data_process.py
mchiusi: I have removed it (unused) and adapt the import lines in `bye_splits/plot/display_plotly/processing.py`

bfonta:
Thanks for the answers above.

> In the display code I remove every "ROI" label. Should I also change the name to task/roi.py?

I suppose you mean changing it to ```task/coarse_seeding.py``` (or similar), and I think that is a good thing to do. Just remember to update all the imports as well.

All the rest is good :+1:

bye_splits/data_handle/data_process.py
mchiusi: Ok I changed a bit the PU variable in `main.py`, `processing.py` and `data_process.py` files.

bye_splits/data_handle/data_process.py
mchiusi: I agree but maybe we should discuss a bit in the next days about this variable.
 -  It can be confused with the same debug variable when running the webApp
 - Which is its purpose? Just creating a different tag in the files' name? We can remove it I think..

bye_splits/data_handle/data_process.py
mchiusi: Is it fine to use the approximation in `common.get_pt()`?

bye_splits/data_handle/data_process.py
mchiusi: Already corrected as said in the comment above

bye_splits/data_handle/data_process.py
bfonta: What do we gain by doing so? Unless a very good reason exists, we should always try to use the variables as close to the simulation as possible.

bye_splits/scripts/run_radii_chain.py
isehle: Indeed, but its functionality since been deprecated and replaced+extended by [run_cluster_size.py](https://github.com/isehle/bye_splits/blob/layerOpt/bye_splits/scripts/cluster_size/run_cluster_size.py).

isehle: @mchiusi, it would probably be better to extend the functionality of the current file than to create a whole new file. We can discuss it whenever you're free.

isehle (commented):
Overall looks like this is going to bring some great new functionality, thanks for the work! Mostly just a couple of minor comments, though I think we should discuss the minor differences between [run_radii_chain](https://github.com/bfonta/bye_splits/blob/714741378683e7c1c8eb83c784d063595bd45375/bye_splits/scripts/run_radii_chain.py) and [run_cluster-size](https://github.com/bfonta/bye_splits/blob/714741378683e7c1c8eb83c784d063595bd45375/bye_splits/scripts/cluster_size/run_cluster_size.py) because they are essentially duplicate files with minor differences. We can likely find a way to extend the functionality of the latter to include the changes included in the former.

bye_splits/data_handle/data_process.py
mchiusi: The only reason I decided to use that function is that we can omit one column from the dataframe, and the stored files are lightweight. I don't know exactly how lightweight, but at least somewhat. But I agree that the pt information should be more precise.

bye_splits/data_handle/data_process.py
mchiusi: Ok, I have removed it and added the possibility to choose debug mode when launching the app (via parser).

bye_splits/data_handle/event.py
mchiusi: Added `try...except` statement

bye_splits/data_handle/geometry.py
mchiusi: This value allows me to distinguish between Si sensors and scintillators. If I remember correctly (please correct me if I'm wrong), in the event files this variable assumes the same value (from CMSSW?). That's why I'm using it [here](https://github.com/bfonta/bye_splits/blob/714741378683e7c1c8eb83c784d063595bd45375/bye_splits/plot/display_plotly/processing.py#L117) in the merging process between geometry and events

bye_splits/plot/display_plotly/plot_event.py
mchiusi: No, this is simply a tradeoff between display speed / file size and how many TCs we want to display around the gen particle. In the case of pileup events, if we increase it to 100, for example, plotly will take many seconds to display the event, and the file will be really huge.

bye_splits/plot/display_plotly/processing.py
mchiusi: I have already remove this dictionary, since I have introduced a second dropdown menu and a second variable (`pu` handling the pileup) in addition to `particles`. I will upload the code today  or tomorrow with all these small comments.

bye_splits/plot/display_plotly/processing.py
mchiusi: Already changed (from comment above)

bye_splits/plot/display_plotly/processing.py
mchiusi: Same comment are before! To ensure that it is a tile and not a silicon module, I need to rely on (at least) one variable between `waferu` and `waferv`, otherwise, it is possible to find a tile and a silicon module that share the same (u,v) coordinate.

bye_splits/plot/display_plotly/processing.py
mchiusi: Already done

bye_splits/tasks/roi.py
mchiusi: Ops I forgot it

config.yaml
mchiusi: We can discuss in the comment above

bye_splits/tasks/cluster.py
```diff
@@ -93,7 +94,7 @@ def cluster(pars, in_seeds, in_tc, out_valid, out_plot, **kw):
         assert len(cols) == res.shape[1]
 
         df = pd.DataFrame(res, columns=cols)
-        assert not df.empty
+        #assert not df.empty
```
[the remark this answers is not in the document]
mchiusi: Yes, there are some cases (just a bunch) in which (when using the minimum clustering radius) there aren't TCs associated to the main seed, which I think is not strange.

bye_splits/scripts/run_radii_chain.py
mchiusi: Ahhh sorry Isaac! Since your code was in a folder and not at the same level as the other run_chain files I totally forgot to merge mine and yours together. We can discuss about it together as soon as I come back to the lab.
I honestly don't know if we should do it here or in a different PR. Touching the `run_radii_chain.py` file is quite messy since it's heavily involved in the reconstruction and storage. I think we have to do it, but it will require some time (we will have to test your code and mine). What do you think?

bye_splits/data_handle/data_process.py
bfonta: I then it is better to add the information back, and keep as close as possible to the C++ code.

bye_splits/tasks/cluster.py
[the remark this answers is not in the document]
bfonta: @mchiusi If this is the case, either remove the commented assert, or replace it by something useful. Following what you said (pseudo-code): ```assert not df.empty and cluster_radius != minimum```.

bye_splits/scripts/run_radii_chain.py
bfonta: In the interest of time, I propose we leave this for another PR, and keep it as-is.

mchiusi: Yes and we should also use the same `run_default_chain.py` for both our scripts (so _at least_ `run_cluster_size.py` and `run_radii_chain.py`). At the moment we are calling all the reco chain steps in each of these scripts, resulting in an annoying repetition of the code.

bye_splits/data_handle/data_process.py
mchiusi: Okok nice! In this case we have to produce the corresponding new files, containing that variable (`produce.py` has been updated in this PR). For the moment we have `/data_CMS/cms/alves/L1HGCAL/skim_with_pt_ntuple_1.root`, which contains 0PU photons, produced by @bfonta.

mchiusi:
> > Thanks for the answers above.
> > > In the display code I remove every "ROI" label. Should I also change the name to task/roi.py?
> > 
> > 
> > I suppose you mean changing it to `task/coarse_seeding.py` (or similar), and I think that is a good thing to do. Just remember to update all the imports as well.
 
I've changed the names of some files and all the variables related to the ROI. Now the two chains are called:
 - *default chain*
 - *new chain*

I have tested both my display app and the `run_meta.py` and they works well. Let me know if there are inconsistencies  in the code due to many changes I have done.
> 
> > All the rest is good 👍

mchiusi:
> ### Third review
> These are almost all very minor comments, and we can merge it as soon as you address them. Could you just provide an answer to previous review comments?

All your comments are now included in the latest commit.

bfonta (approved):

isehle (approved):

PR #47, opened 2023-10-08
New reconstruction chain in the webApp

mchiusi:
In this PR, I have added the possibility to run the **new reconstruction chain** from the 3D display app, allowing users to view its outputs on the screen. Users now have the option to display cluster seeds from the new reconstruction chain, which includes the _new coarse seeding_ and the _new seeding algorithm_, or from the default chain by simply interacting with a switch.

The scripts required to run the new reconstruction chain were already present in the Python framework. When calling for new events from the app, both chains are executed, so that the final user can compare the two chains applied to the same event. The `new_radii_chain` function has been updated accordingly. The dataframes are then processed and stored in the usual hdf5 files, where a new key has been added. This key allows for distinguishing between the new chain and the default chain and can assume the values of 'new' or 'default'.

Here's an example of a photon (200 pileup) processed by both chains. The difference in performance is truly impressive.
image
image

### DIFF
--- a/bye_splits/plot/display_plotly/main.py
+++ b/bye_splits/plot/display_plotly/main.py
@@ -29,7 +29,7 @@ app.layout = html.Div([
                   dbc.Nav([dbc.DropdownMenu([
                                dbc.DropdownMenuItem('3D view', id='3D view'),
                                dbc.DropdownMenuItem('Layer view', id='Layer view')],
-                           label="Pages", nav=True)
+                           label="View", nav=True)
                           ])
               ],brand="3D trigger cells visualization",color="primary",dark=True,)]),
     html.Br(),
@@ -72,22 +72,21 @@ def update_event(particle, pu, n_click, submit_event, event, page):
     button_clicked = ctx.triggered_id
     if button_clicked == 'submit-val':
         if event is None:  raise RuntimeError("Please select manually an event or click on 'Random event'.")
-
+    
     df_dict, event = process.get_data(common.dot_dict(vars(args)), particle, pu, event)
     
     gen_info = next(reversed(df_dict.values()))
-    slider = plt.prepare_slider(next(iter(df_dict.values())), page)
+    slider = plt.prepare_slider(next(iter(df_dict['default'].values())), page)
 
-    if particle == 'photons' and pu == '200 PU':
+    if pu == '200 pileup':
         df_dict = plt.geom_selection(df_dict)
     if particle != 'pions': checkbox = ['Coarse seeding','Layer selection','Seed index']
     else: checkbox = ['Layer selection','Seed index']
 
-    df_dict = {k: v.to_json() for k, v in df_dict.items()}
-    event_selected_message = 'Event {} selected. Gen Particle (η={:.2f}, ϕ={:.2f}), pₜ={:.2f} GeV.'
-    return (event_selected_message.format(int(event), gen_info['gen_eta'].values[0], 
-            gen_info['gen_phi'].values[0], gen_info['gen_pt'].values[0]), 
-            checkbox, slider, df_dict, None)
+    df_dict = {key:val for key, val in df_dict.items() if key != 'gen'}
+    json_df_dict = {chain: {coef: df.to_json() for coef, df in v.items()} for chain, v in df_dict.items()}
+    event_selected_message = 'Event {} selected. Gen Particle (η={:.2f}, ϕ={:.2f}, p$_{{\\text{T}}}$={:.2f} GeV).'
+    return ('Event '+ event +' selected. Gen Particle η={:.2f}, ϕ={:.2f}, '.format(gen_info['gen_eta'].values[0],gen_info['gen_phi'].values[0])+ 'p$_{T}$'+'={:.2f} GeV'.format(gen_info['gen_pt'].values[0]), checkbox, slider, json_df_dict, None)
 
 @app.callback(
     [
@@ -100,13 +99,14 @@ def update_event(particle, pu, n_click, submit_event, event, page):
         Input('slider_cluster', 'value'),
         Input('mip', 'value'),
         Input('checkbox', 'value'),
+        Input('chain', 'value'),
     ],
     [State('page', 'key')]
 )
-def make_graph(data, slider_value, coef, mip, checkbox, page):
+def make_graph(data, slider_value, coef, mip, checkbox, chain, page):
     assert float(mip) >= 0.5, 'mip\u209C value out of range. Minimum value 0.5 !'
-    df_dict = {k: pd.read_json(v) for k, v in data.items()}
-    df = df_dict[str(coef)]
+    df_dict = {k: {coef: pd.read_json(df) for coef, df in v.items()} for k, v in data.items()}
+    df = df_dict[chain][str(coef)]
     df_sel = df[df.tc_mipPt >= mip]
 
     if page == '3D':
@@ -144,4 +144,4 @@ if __name__ == '__main__':
     args = parsing.parser_display_plotly()
     process = processing.Processing() 
     
-    app.run_server(host=args.host, port=args.port, debug=args.debug)
+    app.run_server(host=args.host, port=args.port, debug=True)
--- a/bye_splits/plot/display_plotly/plot_event.py
+++ b/bye_splits/plot/display_plotly/plot_event.py
@@ -14,13 +14,17 @@ from bye_splits.tasks.coarse_seeding import cs_dummy_calculator
 def geom_selection(df_dict):
     gen_info = next(reversed(df_dict.values()))
 
-    radius_gen = 50 # radius to select a specific region around the gen
-    for coef in list(df_dict.keys())[:-1]:
-        eta = gen_info['gen_eta'].values[0]
-        phi = gen_info['gen_phi'].values[0]
-        x_gen, y_gen = sph2cart(eta, phi)
-        df_dict[coef] = df_dict[coef][np.sqrt((x_gen-df_dict[coef].tc_x)**2+(y_gen-df_dict[coef].tc_y)**2)<radius_gen]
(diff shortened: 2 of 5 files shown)


bye_splits/plot/display_plotly/main.py
```diff
@@ -100,13 +101,15 @@ def update_event(particle, pu, n_click, submit_event, event, page):
...
     ],
     [State('page', 'key')]
 )
-def make_graph(data, slider_value, coef, mip, checkbox, page):
+def make_graph(data, slider_value, coef, mip, checkbox, chain, page):
     assert float(mip) >= 0.5, 'mip\u209C value out of range. Minimum value 0.5 !'
-    df_dict = {k: pd.read_json(v) for k, v in data.items()}
-    df = df_dict[str(coef)]
+    df_dict = {k: {coef: pd.read_json(df) for coef, df in v.items()} for k, v in data.items()}
+    chain_str = 'default' if not chain else 'new'
```
bfonta, line 111: You can convert this already to a dictionary of keys, for support future additional chains.

bye_splits/plot/display_plotly/plot_event.py
```diff
@@ -14,13 +14,17 @@
...
 
-    radius_gen = 50 # radius to select a specific region around the gen
-    for coef in list(df_dict.keys())[:-1]:
-        eta = gen_info['gen_eta'].values[0]
-        phi = gen_info['gen_phi'].values[0]
-        x_gen, y_gen = sph2cart(eta, phi)
-        df_dict[coef] = df_dict[coef][np.sqrt((x_gen-df_dict[coef].tc_x)**2+(y_gen-df_dict[coef].tc_y)**2)<radius_gen]
-
+    radius_gen = 40  # radius to select a specific region around the gen
+    for chain in ['default', 'new']:
```
bfonta, line 18: the keys should be stored as a tuple variable somewhere, accessible to all functions

bye_splits/plot/display_plotly/processing.py
```diff
@@ -61,9 +63,11 @@ def get_data(self, pars, particles, pu_str, event = ''):
     def get_event(self, event):
         """   Load dataframes of a particular event from the hdf5 file   """
         dic = {}
-        for coef in self.coefs:
-            key = '/ev_'+str(event)+'/coef_'+str(coef)[2:]
-            dic[coef] = pd.read_hdf(self.filename, key)
+        for chain in self.chain:
+          dic[chain] = {}
+          for coef in self.coefs:
+            key = '/ev_'+str(event)+'/'+chain+'/coef_'+str(coef)[2:]
```
bfonta, line 69: Can you add a comment in the doc string explaining why "[2:]" is needed?

bye_splits/scripts/run_radii_chain.py
```diff
@@ -23,44 +23,85 @@
...
     for coef in coefs:
-        cluster_d["CoeffA"] = [coef, 0] * 50
-        nevents = tasks.cluster.cluster_default(pars, **cluster_d)
-        store_file = pd.HDFStore(common.fill_path(cluster_d["ClusterOutPlot"], **pars), mode='r')
+        cluster_params["CoeffA"] = [coef, 0] * 50
+        if chain == 'default': 
+            nevents = tasks.cluster.cluster_default(pars, **cluster_params)
+            store_file = pd.HDFStore(common.fill_path(cluster_params["ClusterOutPlot"], **pars), mode='r')
+        else:
+            nevents = tasks.cluster.cluster_cs(pars, **cluster_params)
```
bfonta, line 35: How come is it still called "CS"? Can you quickly change it everywhere?

bye_splits/scripts/run_radii_chain.py
```diff
@@ -23,44 +23,85 @@
...
+    
+    return cluster_data, events
+
+def merge_cluster_data_with_event(df_event_tc, cluster_data, index):
+    merged_data = {}
+    
+    for coef in cluster_data.keys():
+        merged_data[coef] = pd.merge(
+            left=cluster_data[coef][index],
+            right=df_event_tc[df_event_tc.tc_layer % 2 != 0],
```
bfonta, line 58: This is only valid for the CE-E.

bye_splits/scripts/run_radii_chain.py
```diff
@@ -23,44 +23,85 @@
...
-        dict_cluster[str(coef)[2:]] = df_list
-        store_file.close() 
 
+        cluster_data[str(coef)[2:]] = df_list
+        store_file.close()
+    
+    return cluster_data, events
+
+def merge_cluster_data_with_event(df_event_tc, cluster_data, index):
+    merged_data = {}
```
bfonta, line 53: Please add a doc string. This function is central to the overall logic.

config.yaml
```diff
@@ -125,15 +125,15 @@ varGeometry:
...
   PU200:
-    photons: 'photons_200PU_cluster.hdf5'
-    electrons: 'electrons_200PU_cluster.hdf5'
-    pions: 'pions_200PU_cluster.hdf5'
+    photons: 'photons_200PU_cluster'
+    electrons: 'electrons_200PU_cluster'
+    pions: 'pions_200PU_cluster'
   reprocess: False
-  coefs: [0.006,0.010,0.014,0.018,0.022,0.026,0.030]
+  coefs: [0.012,0.018,0.024]
```
bfonta, line 136: petty: add spaces between numbers for readability

bye_splits/scripts/run_radii_chain.py
```diff
@@ -23,44 +23,85 @@
...
+
+    # New chain
+    fill_cs_params = params.read_task_params('cs')
+    tasks.coarse_seeding.coarse_seeding(pars, df_gen, df_cl, df_tc, **fill_cs_params)
+
+    seed_cs_params = params.read_task_params('seed_cs')
+    tasks.seed_cs.seed_cs(pars, **seed_cs_params)
+
+    cluster_params = params.read_task_params("cluster")
+    cluster_data_new, events = get_cluster_data(cluster_params, 'new', coefs, pars)
```
bfonta, line 89: Can you convert the two independent chains to a loop on "default" and "new" (after changing the latter label)?

bfonta (commented):
Great! General comments:

- the key "new" for the new chain is not ideal, because as soon as we add a third chain (or a small variation of one of the current two chains) it will have to be changed
- following the same reasoning, it is likely better to use a widget with more than two options

bye_splits/plot/display_plotly/main.py
mchiusi: Sure

bye_splits/plot/display_plotly/processing.py
mchiusi: Yess

bye_splits/scripts/run_radii_chain.py
mchiusi: Is not the correct name? I'm confused..

bfonta: You're right, I thought we had resort to call it "ROI", but it is actually the opposite. Sorry.

bye_splits/scripts/run_radii_chain.py
mchiusi: Corrected, now using the `disconnectedTriggerLayers` variable in the `config.yaml` file.

config.yaml
mchiusi: Ok

bye_splits/plot/display_plotly/plot_event.py
mchiusi: Yes sorry! You are definitely right! I can just read the key directly from the dictionary :)

mchiusi:
> Great! General comments:
> 
> * the key "new" for the new chain is not ideal, because as soon as we add a third chain (or a small variation of one of the current two chains) it will have to be changed
> * following the same reasoning, it is likely better to use a widget with more than two options

I've added a dropdown menu instead of a switch so that we can easily add multiple chains in the code. Now the _new_ chain is called _coarse seeding_ chain.

bye_splits/scripts/run_radii_chain.py
bfonta: Why can't you do it already?

mchiusi: Because I would like to wait and discuss it with you @isehle as well, we could open a new PR. This would involve merging my `run_radii_chain` and Isaac's `run_cluster_size` scripts to use the `run_default_chain` and `run_new_chain`. It's not a trivial task because the current code in these latter scripts doesn't allow running the clustering step multiple times. While I don't think it's difficult, I would prefer to open a new PR and address these code repetitions. After making these modifications, we should also test all the chains independently from the webApp. What do you think? We can discuss this further tomorrow :)

bfonta: I agree, but would add the discussion above to a new issue so that we keep track of things. Can you please do it?

bfonta:
Additional minor comments:
- Could you please replace "t" by "T" when showing the transverse momentum?
- Why are eta and phi within parenthesis and not pT? Can you uniformise?
- I think you could add the "Select a particular clustering radius: " slider to the right of "Select manually an event:", so to gain vertical space.

mchiusi:
> Additional minor comments:
> 
> * Could you please replace "t" by "T" when showing the transverse momentum?
> * Same for "Threshold in [mipₜ]", where you can also remove the "[" and "]"
> * Why are eta and phi within parenthesis and not pT? Can you uniformise?
> * I think you could add the "Select a particular clustering radius: " slider to the right of "Select manually an event:", so to gain vertical space.
> * Replace "Pages" by "View", or similar, since it is IMO clear
> 
> Once this is addressed I am ready to approve. What about @isehle?

Implementations done! Now the page layout is a bit better!

bfonta (approved):
All set. Thank you for all the work, and for addressing the (many) comments.

PR #48, opened 2023-10-09
Job submission

isehle:
### PR Description

This PR is specifically for submitting jobs on HT Condor, using the new Python script ```job_submit.py``` located in ```bye_splits/production/submit_scripts```. All configuration is done in ```config.yaml``` under ```job```. This configuration should contain the usual Condor related variables (user, proxy, queue, local) as well as a path to the script you wish to run, and a list of arguments that the script accepts. There are then configurations for each particle type listed under their names (plural), which should contain:
1) ```submit_dir```, i.e. the path where you want your submission related scripts to be written to and read from;
2) ```files``` which should contain a path to a ```txt``` file with your various parameters, line delimited;
3) ```files_per_batch``` which is self explanatory.

In the case of the cluster size studies, ```files``` points to a ```.txt``` file with all clustering radii, line delimited. 
```job_submit.py``` contains three classes, ```JobBatches``` which splits the total number of jobs into batches as defined above. If the chosen number can't be split evenly, the last batch will be shorter than the preceding ones. The second class, ```CondJobBase``` writes the individual batch files if needed, and writes the ```.sh``` and ```.sub``` scripts needed for the job(s). Finally, ```CondJob``` creates ```configs/```, ```jobs/```,  and ```logs/``` sub-directories under the ```submit_dir``` in the configuration script, then creates and runs the jobs. The base name for all files written by ```job_submit.py``` will be the same as the basename for the ```script``` being called, as set in the configuration. The file information is stored in buffer-lists before being written, and are checked against files with the same name before being written. If the contents of the new file are equivalent to a previously written file, the script will opt to use that file rather than write a new one. Otherwise, it will increment the version number, write the new file and use that. This is handled through a few new functions in ```common.by```, with the main function being ```conditional_write```.

### How has this been tested?

Tested against all particle types with both 0 and 200 PU.

### Checklist

### CODE REVIEW

**Review by bfonta — COMMENTED** [review]
Nice to have this close to integration! A few general comments:

+ IIRC ```files``` was a bad argument name that had to be changed
+ Instead of handling the submission with files, cluttering the environment, it would be cleaner to use other options for the ```queue ...``` command (see [the docs](https://htcondor.readthedocs.io/en/latest/users-manual/submitting-a-job.html#submitting-many-similar-jobs-with-one-queue-command))
+ Could you please remove the ```produce.cc``` file, which is no longer needed?
+ You are missing doc strings everywhere

File: bye_splits/production/submit_scripts/job_submit.py [hunk 1]
```diff
@@ -0,0 +1,222 @@
...
+particle_var = lambda part, var: config["job"][part][var]
+
+class JobBatches:
+    def __init__(self, particle, config):
+        self.particle = particle
+        self.config = config
+
+    def setup_batches(self):
+
+        my_batches = lambda files, files_per_batch: [files[i: i + files_per_batch] for i in range(0, len(files), files_per_batch)]
```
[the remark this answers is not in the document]
isehle: Good catch, it comes from an older iteration where I was using the function a few times, but the variable is declared directly now.

bye_splits/production/submit_scripts/job_submit.py
```diff
@@ -0,0 +1,222 @@
...
+            part_submit_dir = particle_var(self.particle, "submit_dir") + "ntuples/"
+            paths = [
+                "{}{}".format(part_submit_dir, file) for file in os.listdir(part_submit_dir) if file.startswith("ntuple")
+            ]
+        
+        batches = my_batches(paths, files_per_batch)
+
+        return batches
+    
+class CondJobBase(JobBatches):
```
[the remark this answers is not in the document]
isehle: Yes that's better, changed now!

bye_splits/production/submit_scripts/job_submit.py
```diff
@@ -0,0 +1,222 @@
...
+            current_version.append("bash {}".format(self.script))
+            
+        # Write the file only if an identical file doesn't already exist
+        global sub_file
+        sub_file = common.conditional_write(submit_file_versions, submit_file_name_template, current_version)
+
+    def prepare_multi_job_condor(self):
+        log_dir = "{}logs/".format(self.particle_dir)
+        
+        batch_files = current_batch_versions
```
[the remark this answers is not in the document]
isehle: Indeed seems superfluous; the latest commit removes it.

bye_splits/production/submit_scripts/job_submit.py
```diff
@@ -0,0 +1,222 @@
...
+
+        arg_dict = {}
+        for arg in self.args:
+            if arg=="filename":
+                arg_dict[arg] = batch_files
+            elif arg=="particles":
+                arg_dict[arg] = self.particle
+            elif arg=="pileup":
+                arg_dict[arg] = "PU0" if "PU0" in batch_files[0] else "PU200"
```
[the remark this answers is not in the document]
isehle: Arguments are handled with a `key/value` pairing in the latest commit so it's no longer hardcoded.

bye_splits/scripts/cluster_size/condor/run_cluster.py
```diff
@@ -0,0 +1,69 @@
...
+        cluster_d[key] =  "{}_{}_{}_posEta_9oct".format(particles, pileup, name)
+    
+    nevents_end = tasks.cluster.cluster_default(pars, **cluster_d)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="")
+    parser.add_argument("--coef", help="Coefficient to use as the max cluster radius", required=True, type=float)
+    parser.add_argument("--particles", choices=("photons", "electrons", "pions"), required=True)
+    parser.add_argument("--pileup", help="tag for PU200 vs PU0", choices=("PU0", "PU200"), required=True)
+    parser.add_argument("--weighted", help="Apply pre-calculated layer weights", default=False)
```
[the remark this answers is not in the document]
isehle: Indeed I had tried it with a flag before, but was having weird bugs with the parser when I tried that. I seem to have fixed whatever the problem was though, because it's now working with a flag.

config.yaml
```diff
@@ -135,17 +135,40 @@ varGeometry:
   reprocess: False
   coefs: [0.006,0.010,0.014,0.018,0.022,0.026,0.030]
 
+job:
+  user: iehle
+  proxy: ~/.t3/proxy.cert
+  queue: short
+  local: False
+  script: /grid_mnt/vol_home/llr/cms/ehle/NewRepos/bye_splits/bye_splits/production/submit_scripts/condor_cluster_size.sh
+  arguments: [filename, particles, pileup]
```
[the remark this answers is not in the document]
isehle: Updated, structure now follows:
```
job: 
    ...
    iterOver: iter_key
    arguments:
        arg_key_1: val_1
        ...
    particles:
        - photons:
        - pions:
        - ...
```

where the ```iter_key``` should correspond to the ```arg_key``` continuing the values that the script will iterate over.

bye_splits/production/submit_scripts/job_submit.py
```diff
@@ -0,0 +1,222 @@
...
+parent_dir = os.path.abspath(__file__ + 5 * "../")
+sys.path.insert(0, parent_dir)
+
+from bye_splits.utils import params, common
+
+import subprocess
+import yaml
+
+# Read particle specific variables from the YAML file
+particle_var = lambda part, var: config["job"][part][var]
```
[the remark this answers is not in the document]
isehle: No good reason, it's moved inside now.

bye_splits/production/submit_scripts/job_submit.py
```diff
@@ -0,0 +1,222 @@
...
+        sub_dir = "{}subs/".format(self.particle_dir)
+
+        if not os.path.exists(sub_dir):
+            os.makedirs(sub_dir)
+        script_basename = os.path.basename(self.script).replace(".sh", "").replace(".py", "")
+
+        submit_file_name_template = "{}{}_submit.sh".format(sub_dir, script_basename)
+        submit_file_versions = common.grab_most_recent(submit_file_name_template, return_all=True)
+
+        current_version = []
```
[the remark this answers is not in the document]
isehle: It's simply the easiest way I could find to implement the `conditional_write()` procedure, because a previously written file can be read into a list with ```contents = file.readlines()``` and from there it's easy to check the content equivalence with ```contents == current_version```. See [this function](https://github.com/isehle/bye_splits/blob/b894653abbbcda8f12512b11b7d54b1d143ab90c/bye_splits/utils/common.py#L185) used by `conditional_write()`.

bye_splits/scripts/cluster_size/run_combine.py
```diff
@@ -0,0 +1,137 @@
...
+    original_cols = [col.replace("weighted_","") for col in weighted_cols]
+    weighted_cols += ["event"]
+    original_cols += ["event"]
+
+    original_df = cl_df[original_cols]
+    weighted_df = cl_df[weighted_cols].rename(dict(zip(weighted_cols, original_cols)), axis=1)
+
+    return original_df, weighted_df
+
+def normalize_df(cl_df, gen_df, dRThresh):
```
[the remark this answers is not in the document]
isehle: Indeed, but it's difficult to think of a succinct name that's better suited to it. For now I've switched it to `combine_normalize` but if you can think of a better one, I'm all ears.

bye_splits/scripts/cluster_size/run_init_tasks.py
```diff
@@ -0,0 +1,83 @@
...
+    df_tc_pos, df_tc_neg = df_tc[ df_tc.tc_eta > 0 ], df_tc[ df_tc.tc_eta < 0]
+
+    for i in range(2):
+        if i==1:
+            df_gen, df_cl, df_tc = df_gen_neg, df_cl_neg, df_tc_neg
+            eta_tag = "negEta"
+            df_gen["gen_eta"], df_cl["cl3d_eta"], df_tc["tc_eta"], df_tc["tc_z"] = abs(df_gen.gen_eta), abs(df_cl.cl3d_eta), abs(df_tc.tc_eta), abs(df_tc.tc_z)
+        else:
+            df_gen, df_cl, df_tc = df_gen_pos, df_cl_pos, df_tc_pos
+            eta_tag = "posEta"
```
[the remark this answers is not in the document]
isehle: Yeah this was ugly; updated to a dictionary now.

config.yaml
```diff
@@ -72,7 +72,7 @@ varEvents:
     geta: 'good_genpart_exeta'
     gphi: 'good_genpart_exphi'
     gen: 'good_genpart_energy'
-    gpt: 'good_genpart_pt'
+#    gpt: 'good_genpart_pt'
```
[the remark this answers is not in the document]
isehle: Fair, but I'll have to produce new skims that contain this column, because the current versions which contain both negative and positive eta only have the energy. Not a problem though, it's uncommented now.

bye_splits/scripts/cluster_size/run_combine.py
```diff
@@ -0,0 +1,137 @@
...
+sys.path.insert(0, parent_dir)
+
+from utils import params, common
+
+from data_handle.data_process import get_data_reco_chain_start
+
+import random
+import re
+
+random.seed(10)
```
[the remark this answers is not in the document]
isehle: This is actually a holdover from a previous use, it's not used here anyway, so I've removed it.

isehle:
The two commits that I just pushed should address all of the comments, let me know if there's still something I can update.

bye_splits/scripts/cluster_size/run_init_tasks.py
```diff
@@ -0,0 +1,86 @@
...
+import sys
+
+import yaml
+
+def start_chain(pars, cfg):
+    """Runs the first three steps of the TPG on
+    negative and positive eta samples."""
+
+    particles = cfg["selection"]["particles"]
+    pileup    = "PU0" if not cfg["clusterStudies"]["pileup"] else "PU200"
```
bfonta, line 28: as in the previous review, the pile-up can take many values

bye_splits/utils/cl_helpers.py
```diff
@@ -36,17 +30,12 @@ def get_last_version(name):
         )
```
bfonta, line 30: IIUC, the replace is not needed if you specify the extension in ```pattern```.

bye_splits/utils/cl_helpers.py
```diff
@@ -36,17 +30,12 @@ def get_last_version(name):
...
     version = 0 if not os.path.exists(name) else get_last_version(name)
     return f"{base}_v{str(version+1)}{ext}"
 
-
-# def_k = 0.0
-
-
 def closest(list, k=0.0):
     """Find the element of a list containing strings ['coef_{float_1}', 'coef_{float_2}', ...] which is closest to some float_i"""
     try:
```
bfonta, line 41: not clear what is happening when ```ValueError``` is thrown. Can you make it more explicit with the code?

bye_splits/utils/cl_helpers.py
```diff
@@ -154,44 +143,24 @@ def get_input_files(base_path, pile_up=False):
...
+        weights_by_radius = {}
+        for file in files:
+            radius = float(file.replace(".hdf5","").replace(f"{basename}_","").replace("r0","0").replace("p","."))
+            infile = particle_dir+file
+            with pd.HDFStore(infile, "r") as optWeights:
+                weights_by_radius[radius] = optWeights[mode]
+    
+        weights_by_particle[particle] = weights_by_radius
+    
+    weights_by_particle["electrons"] = weights_by_particle["photons"]
```
bfonta, line 164: can you add a comment for clarification?

bye_splits/utils/common.py
```diff
@@ -16,6 +16,8 @@
 import numpy as np
 import pandas as pd
 
+import re
```
bfonta, line 19: You don't use it.

bye_splits/utils/job_helpers.py
```diff
@@ -0,0 +1,81 @@
...
+    dir, file = os.path.split(file_path)
+    base, ext = os.path.splitext(file)
+    i = 0
+    file_path = "{}/{}_v{}{}".format(dir, base, i, ext)
+    while os.path.exists(file_path):
+        i += 1
+        file_path = "{}/{}_v{}{}".format(dir, base, i, ext)
+    return file_path
+
+# Grab the most recent version of the file corresponding to the template file_path (or return all matches)
```
bfonta, line 24: In a doc-string, not an external comment.

bye_splits/utils/job_helpers.py
```diff
@@ -0,0 +1,81 @@
...
+def grab_most_recent(file_path, return_all=False):
+    dir, file = os.path.split(file_path)
+    base, ext = os.path.splitext(file)
+    files = os.listdir(dir)
+    version_pattern = re.compile("{}_v(\\d+)\\{}".format(base, ext))
+    matches = [version_pattern.search(file) for file in files]
+    matches = [match for match in matches if not match is None]
+    if len(matches) > 0:
```
bfonta, line 32: The function does not return if the if condition is not met.

bye_splits/utils/job_helpers.py
```diff
@@ -0,0 +1,81 @@
...
+
+def conditional_write(file_versions, file_template, current_version):
+    """
+    Loops through the files in <file_versions>, comparing their contents
+    to the current version. If an identical version is found, the function
+    breaks and does nothing. Otherwise, it will write the contents in
+    <current_version> to an updated version number whose basename corresponds to
+    <file_template>.
+    """
+    if file_versions != None:
```
bfonta, line 67: ```is not None```

config.yaml
```diff
@@ -135,17 +135,44 @@ varGeometry:
...
+  local: False
+  script: /grid_mnt/vol_home/llr/cms/ehle/NewRepos/bye_splits/bye_splits/production/submit_scripts/condor_cluster_size.sh
+  iterOver: radius
+  arguments:
+    radius: [0.001, 0.002, 0.003, 0.004, 0.005, 0.006, 0.007, 0.008, 0.009,
+            0.01 , 0.011, 0.012, 0.013, 0.014, 0.015, 0.016, 0.017, 0.018,
+            0.019, 0.02 , 0.021, 0.022, 0.023, 0.024, 0.025, 0.026, 0.027,
+            0.028, 0.029, 0.03 , 0.031, 0.032, 0.033, 0.034, 0.035, 0.036,
+            0.037, 0.038, 0.039, 0.04 , 0.041, 0.042, 0.043, 0.044, 0.045,
+            0.046, 0.047, 0.048, 0.049, 0.05 ]
```
bfonta, line 151: petty: remove last space

config.yaml
```diff
@@ -135,17 +135,44 @@ varGeometry:
...
+  electrons:
+    submit_dir: /data_CMS/cms/ehle/L1HGCAL/PU0/electrons/
+    args_per_batch: 10
+  pions:
+    submit_dir: /data_CMS/cms/ehle/L1HGCAL/PU0/pions/
+    args_per_batch: 10
+
 clusterStudies:
   localDir: /home/llr/cms/mchiusi/event_display/bye_splits/data/new_algos/
   ehleDir: /eos/user/i/iehle/
```
bfonta, line 166: Maybe a less user-dependent variable name, to make it clear what it is used for? ;) ```localDir``` is quite clear.

bfonta (commented):
I've added a few more comments.
Regarding the previous review, you did not address the general comments, and importantly the comment on the ```queue```-related command, which is quite central to the PR. It would require a bit of extra work, but would be more robust for future changes and more readable. It would make the code shorter, since IIUC much of the code for checking the files would not be required any longer.

And thanks for documenting the code; it makes a big difference!

bye_splits/scripts/cluster_size/run_init_tasks.py
```diff
@@ -0,0 +1,86 @@
...
+    df_gen_pos, df_gen_neg = df_gen[ df_gen.gen_eta > 0 ], df_gen[ df_gen.gen_eta < 0 ]
+    df_cl_pos, df_cl_neg   = df_cl[ df_cl.cl3d_eta > 0 ],  df_cl[ df_cl.cl3d_eta < 0 ]
+    df_tc_pos, df_tc_neg   = df_tc[ df_tc.tc_eta > 0 ],    df_tc[ df_tc.tc_eta < 0 ]
+
+    eta_dict = {"negEta": {"df_gen": df_gen_neg,
+                           "df_cl": df_cl_neg,
+                           "df_tc": df_tc_neg},
+                "posEta": {"df_gen": df_gen_pos,
+                           "df_cl": df_cl_pos,
+                           "df_tc": df_tc_pos}
```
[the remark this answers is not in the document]
isehle: The keys are already aligned, do you mean to align the colons+values? Fine for the name change.

bye_splits/scripts/cluster_size/run_init_tasks.py
```diff
@@ -0,0 +1,86 @@
...
+                "posEta": {"df_gen": df_gen_pos,
+                           "df_cl": df_cl_pos,
+                           "df_tc": df_tc_pos}  
+                }
+
+    for eta_tag, dfs in eta_dict.items():
+        df_gen, df_cl, df_tc = dfs.values()
+
+        if eta_tag=="negEta":
+            df_gen["gen_eta"], df_cl["cl3d_eta"], df_tc["tc_eta"], df_tc["tc_z"] = abs(df_gen.gen_eta), abs(df_cl.cl3d_eta), abs(df_tc.tc_eta), abs(df_tc.tc_z)
```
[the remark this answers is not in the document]
isehle: Indeed, the logic doesn't support negative values for eta. I don't remember exactly where it breaks, but yes, that's why I worked with absolute values.

bye_splits/scripts/cluster_size/run_combine.py
```diff
@@ -0,0 +1,147 @@
...
+    return original_df, weighted_df
+
+def combine_normalize(cl_df, gen_df, dRThresh):
+    """Combines a cluster dataframe with an associated
+    gen-level dataframe. Useful variables that may or may not
+    be present are added to the combined dataframe before
+    normalized energy and pt columns are added."""
+
+    cl_df=cl_df.reset_index().set_index(["event","seed_idx"])
+    combined_df = cl_df.join(
```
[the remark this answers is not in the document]
isehle: I personally find the longer name to be more readable than shorter names. I can change it if you really think it's important, but I would prefer not to.

bye_splits/production/submit_scripts/job_submit.py
```diff
@@ -0,0 +1,220 @@
...
+            arg_vals = [self.args[key] for key in sub_args]
+            all_vals = ["{}".format(batch).replace(", ", ";")]+arg_vals
+            all_vals = ", ".join(all_vals) + "\n"
+            current_version.append(all_vals)
+
+        current_version.append(")")
+
+    def prepare_batch_submission(self):
+        """Writes the .sh script that constitutes the executable
+        in the .sub script. The basename will be the same as the fundamental
```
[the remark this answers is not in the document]
isehle: Updated doc string:
```
"""Writes the .sh script that constitutes the executable
        in the .sub script. The basename will be the same as the running script, i.e.
        the script set in the configuration file. This is then followed by a version number.
        Stores the contents in a list that's used as a buffer and checked against the content
        in previous versions, only writing the file if an identical file doesn't already exist.
        The version number will be incrimented in this case."""
```

bye_splits/scripts/cluster_size/run_combine.py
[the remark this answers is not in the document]
bfonta: As you wish, no strong preference.

bye_splits/scripts/cluster_size/run_combine.py
```diff
@@ -0,0 +1,147 @@
...
+
+    return combined_df
+
+def combine_files_by_coef(in_dir, file_pattern):
+    """Combines .hdf5 cluster files of individual
+    radii into one .hdf5 file containing dataframes
+    for all radii."""
+
+    files = [
+        file for file in os.listdir(in_dir) if re.search(file_pattern, file) != None and "valid" not in file
```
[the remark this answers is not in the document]
isehle: Rather `is not` but changed now.

bye_splits/scripts/cluster_size/run_combine.py
```diff
@@ -0,0 +1,147 @@
...
+    parser = argparse.ArgumentParser(description="")
+    parser.add_argument("--file", type=str)
+    parser.add_argument("--unweighted", action="store_true")
+    parsing.add_parameters(parser)
+
+    FLAGS = parser.parse_args()
+    pars = common.dot_dict(vars(FLAGS))
+
+    with open(params.CfgPath, "r") as afile:
+        cfg = yaml.safe_load(afile)
```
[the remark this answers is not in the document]
isehle: You have already been using `safe_load()`, see for example [here](https://github.com/bfonta/bye_splits/blob/156fa37eb4aebe00ce65cd6b9bb7517091c88431/bye_splits/data_handle/event.py#L28). A quick search shows that there are no instances of `yaml.load()` in the repository.

bye_splits/utils/cl_helpers.py
isehle: Added this:
```
 '''Weights are calculated from pt_norm distributions, which
    are distorted by brem events for electrons. As this is
    a physics effect uncorrelated to the TPG response, we correct
    electrons with weights derived from photon pt_norm distributions'''
    ```

File: config.yaml (hunk 22 above)
Comment (isehle) [reply config.yaml:166]: It's my name for a reason, though this will soon be depricated by the upcoming dashApp update. See for example [this block](https://github.com/isehle/bye_splits/blob/29a9a85f4837370375b26047dc0aec4709cfa2ab/bye_splits/plot/display_clusters/pages/eff_page.py#L223C1-L230C79) in the `eff_page`. It takes quite a while for the code to process the input files the first time it's run, but then it saves the processed data to an output file that can be read quickly. To save time, these pages first check if the file exists for the existing user, and if not, it checks if it exists in my directory (since I'm more likely to have processed more data than another user). It only reprocesses the data if the file doesn't exist in either place.
That being said, this is obviously a messy way of handling the problem, and the update to the app removes it entirely because the data is processed more quickly from the outset, no longer requiring the caching procedure.

File: bye_splits/utils/cl_helpers.py (hunk 15 above)
Comment (isehle) [reply bye_splits/utils/cl_helpers.py:41]: The `try/except` block is depricated, it was used to handle the `keys()` of two different file versions, one of which needed to be reshaped as above. That's no longer the case, so it's been removed.

File: bye_splits/production/submit_scripts/job_submit.py (hunk 8 above)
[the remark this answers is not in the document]
Comment (isehle) [reply bye_splits/production/submit_scripts/job_submit.py:110]: I don't see how using a different form of `queue` would change the required logic for file parsing/handling. To be more explicit, `prepare_multi_job_condor()` will write a `.sub` file like this:

```
executable = /data_CMS/cms/ehle/L1HGCAL/PU0/photons/subs/condor_cluster_size_submit_v1.sh
Universe = vanilla
Arguments = $(radius) $(particles) $(pileup)
output = /data_CMS/cms/ehle/L1HGCAL/PU0/photons/logs/condor_cluster_size_C$(Cluster)P$(Process).out
error = /data_CMS/cms/ehle/L1HGCAL/PU0/photons/logs/condor_cluster_size_C$(Cluster)P$(Process).err
log = /data_CMS/cms/ehle/L1HGCAL/PU0/photons/logs/condor_cluster_size_C$(Cluster)P$(Process).log
getenv  = true
T3Queue = short
WNTag = el7
+SingularityCmd = ""
include: /opt/exp_soft/cms/t3/t3queue |
queue radius, particles, pileup from (
[0.001;0.002;0.003;0.004;0.005;0.006;0.007;0.008;0.009;0.01], pions, PU0
[0.011;0.012;0.013;0.014;0.015;0.016;0.017;0.018;0.019;0.02], pions, PU0
[0.021;0.022;0.023;0.024;0.025;0.026;0.027;0.028;0.029;0.03], pions, PU0
[0.031;0.032;0.033;0.034;0.035;0.036;0.037;0.038;0.039;0.04], pions, PU0
[0.041;0.042;0.043;0.044;0.045;0.046;0.047;0.048;0.049;0.05], pions, PU0
)
```

As far as I understand, changing  the form of `queue` can potentially reduce complexity in the last part, but that isn't the only part that can change. Most notably, running a different script that accepts different arguments would also change the values of `executable`, `Arguments`, `output`, `error`, and `log`, and `T3Queue` can be either `short` or `long` depending on what you set in `config.yaml`. The logic that's ultimately used by `conditional_write()` handles all of this, and I don't see how a different form of `queue` would be able to do that.

File: bye_splits/production/submit_scripts/job_submit.py [hunk 29]
```diff
@@ -0,0 +1,220 @@
...
+        self.batches      = self.batch.setup_batches()
+
+    def _write_arg_keys(self, current_version):
+        """Writes the line containing the argument
+        names to the buffer list."""
+        
+        arg_keys = ["Arguments ="]
+        for arg in self.args.keys():
+            arg_keys.append("$({})".format(arg))
+        arg_keys = " ".join(arg_keys) + "\n"
```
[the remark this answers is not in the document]
isehle: If you're asking about `arg_keys`, I suppose I don't—changed now. If you're referring to `current_version`, see my previous comment about file comparison with `readlines()`.

bye_splits/utils/cl_helpers.py
isehle: The extension is already specified in `pattern`, do you mean to remove it?

config.yaml
bfonta: Fine, but please link an issue to this, otherwise we will forget about it.

README.md
```diff
@@ -99,6 +100,11 @@ This framework relies on photon-, electron- and pion-gun samples produced via CR
...
 
+<a id="job-submission"></a>
+## Job Submission
+
+Job submission to HT Condor is handled through `bye_splits/production/submit_scripts/job_submit.py` using the `job` section of `config.yaml` for its configuration. The configuration should include usual condor variables, i.e `user`, `proxy`, `queue`, and `local`, as well as a path to the `script` you would like to run on condor. The `arguments` sub-section should contain `key/value` pairs matching the expected arguments that `script` accepts. The variable that you would like to iterate over should be set in `iterOver` and its value should correspond to a `key` in the `arguments` sub-section whose value is a list containing the values the script should iterate over. It then contains a section for each particle type which should contain a `submit_dir`, i.e. the directory in which to read and write submission related files, and `args_per_batch` which can be any number between 1 and `len(arguments[<iterOver>])`.
+
```
bfonta, line 107: I think a working example would be helpful. Could you add it, maybe copying directly from the ```.sub``` files you already have (in a README code block)?

bye_splits/production/submit_scripts/condor_cluster_size.sh
```diff
@@ -0,0 +1,31 @@
...
+while [[ "$#" -gt 0 ]]; do
+    case "$1" in
+        --radius)
+            IFS=";" read -ra radius <<< "${2:1:-1}"
+            shift 2
+            ;;
+        --particles)
+            particles="$2"
+            shift 2
+            ;;
```
bfonta, line 18: petty: you can place the semicolons in the line above

bye_splits/production/submit_scripts/condor_cluster_size.sh
```diff
@@ -0,0 +1,31 @@
...
+        --pileup)
+            pileup="$2"
+            shift 2
+            ;;
```
bfonta, line 22: petty: you can place the semicolons in the line above

bye_splits/production/submit_scripts/condor_cluster_size.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+cd /home/llr/cms/${USER}/NewRepos/bye_splits/bye_splits/scripts/cluster_size/condor/
```
bfonta, line 3: I just realized ```NewRepos``` is specific to you. I suppose you can use ${HOME} or similar to make it user-transparent.

bye_splits/scripts/cluster_size/condor/run_cluster.py
```diff
@@ -0,0 +1,71 @@
...
+    particles = pars["particles"]
+    pileup = pars["pileup"]
+    radius = pars["radius"]
+
+    cl_size_radius = "{}_radius_{}".format(
+        cfg["clusterStudies"]["clusterSizeBaseName"],
+        str(round(radius, 3)).replace(".", "p"),
+    )
+    cluster_d["ClusterOutPlot"], cluster_d["ClusterOutValidation"] = cl_size_radius, cl_size_radius+"_valid"
+    cluster_d["CoeffA"] = [radius] * 50
```
bfonta, line 36: Can you clarify the "50"?

bye_splits/scripts/cluster_size/run_init_tasks.py
```diff
@@ -0,0 +1,83 @@
+# coding: utf-8
+
+_all_ = []
```
bfonta, line 3: Did you forget to add functions here?

bye_splits/utils/cl_helpers.py
```diff
@@ -154,44 +138,28 @@ def get_input_files(base_path, pile_up=False):
...
-
-        # Get rid of duplicates that the dictionary filling produced
-        for key in output_files.keys():
-            output_files[key] = list(set(output_files[key]))
-
-    return output_files
+def read_weights(dir, cfg, version="final", mode="weights"):
+    weights_by_particle = {}
+    for particle in ("photons", "pions"):
+        basename = "optimization_selectOneEffRms_maxSeed_bc_stc" if particle == "pions" else "optimization_selectOneStd_adjustMaxWeight_maxSeed"
```
bfonta, line 144: maybe a dict with the particle type as key would be better in case other particles are later added

bye_splits/utils/cl_helpers.py
```diff
@@ -154,44 +138,28 @@ def get_input_files(base_path, pile_up=False):
...
+        
+        version_dir = "{}/".format(version)
+        particle_dir = "{}{}/optimization/official/{}".format(dir, particle, version_dir)
```
bfonta, line 147: ```os.path.join()``` is slightly better

bfonta (commented):
Only minor details. In addition, could you comment on the following:

> IRC files was a bad argument name that had to be changed

If possible, could @mchiusi also have a careful look?

bye_splits/scripts/cluster_size/condor/run_cluster.py
isehle: Just added a comment, it's the radii in each of the 50 HGCAL layers (in the geometry we're using).

bfonta: Then please use [these](https://github.com/bfonta/bye_splits/blob/main/config.yaml#L51-L52).

README.md
```diff
@@ -99,6 +100,11 @@ This framework relies on photon-, electron- and pion-gun samples produced via CR
...
 
+<a id="job-submission"></a>
+## Job Submission
+
+Job submission to HT Condor is handled through `bye_splits/production/submit_scripts/job_submit.py` using the `job` section of `config.yaml` for its configuration. The configuration should include usual condor variables, i.e `user`, `proxy`, `queue`, and `local`, as well as a path to the `script` you would like to run on condor. The `arguments` sub-section should contain `key/value` pairs matching the expected arguments that `script` accepts. The variable that you would like to iterate over should be set in `iterOver` and its value should correspond to a `key` in the `arguments` sub-section whose value is a list containing the values the script should iterate over. It then contains a section for each particle type which should contain a `submit_dir`, i.e. the directory in which to read and write submission related files, and `args_per_batch` which can be any number between 1 and `len(arguments[<iterOver>])`.
```
mchiusi, line 106: Can you provide and example with a command that you could use to launch the Job submission?

README.md
```diff
@@ -99,6 +100,11 @@ This framework relies on photon-, electron- and pion-gun samples produced via CR
 
 The `PU0` files above were merged and are stored under `/data_CMS/cms/alves/L1HGCAL/`, accessible to LLR users and under `/eos/user/b/bfontana/FPGAs/new_algos/`, accessible to all lxplus and LLR users. The latter is used since it is well interfaced with CERN services. The `PU200` files were merged and stored under `/eos/user/i/iehle/data/PU200/<particle>/`.
 
+<a id="job-submission"></a>
+## Job Submission
```
mchiusi, line 104: You could say explicitly that the **Job submission** can be use for multiple purposes, and it can be a good alternative to the local Skimming procedure presented in this README, just a few lines above.

About this, I was trying to run `produce.py`, but I got the following error while trying to read the input file for the skimming, specified in the `config.yaml` file:
```
Error in <TFile::TFile>: file /eos/user/b/bfontana/FPGAs/new_algos/photons_0PU_bc_stc_hadd.root does not exist
```
Everything is good if I try to run it in local. Do you know if we should add some additional configuration options to avoid this kind of access problems?

mchiusi (changes requested):
Great job, Isaac! 
I've been testing the flexibility of your code by running another script, `produce.py`, and I've identified some critical points that need to be improved. For the rest, everything is good and well-written.

bye_splits/utils/job_helpers.py
```diff
@@ -0,0 +1,83 @@
...
+    if len(matches) > 0:
+        matches = [int(match.group(1)) for match in matches]
+        most_recent = max(matches)
+        if not return_all:
+            file_path = dir + "/" + base + "_v" + str(most_recent) + ext
+        else:
+            file_path = [dir + "/" + base + "_v" + str(f) + ext for f in matches]
+        return file_path
+    else:
+        raise ValueError("There are no versions of the passed file: {}".format(file_path))
```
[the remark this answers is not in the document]
isehle: Yes thank you again for this catch!

bye_splits/utils/job_helpers.py
```diff
@@ -0,0 +1,83 @@
...
+
+def conditional_write(file_versions, file_template, current_version):
+    """
+    Loops through the files in <file_versions>, comparing their contents
+    to the current version. If an identical version is found, the function
+    breaks and does nothing. Otherwise, it will write the contents in
+    <current_version> to an updated version number whose basename corresponds to
+    <file_template>.
+    """
+    if file_versions is not None:
```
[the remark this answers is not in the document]
isehle: I just realized that this is how the `v0` version was supposed to be written. `file_versions` is what `grab_most_recent` returns with `return_all==True`; see for example [`submit_file_versions`](https://github.com/isehle/bye_splits/blob/482e84cd1bab79fa1fe7ce6f4a1f82aff869c3b2/bye_splits/production/submit_scripts/job_submit.py#L109C14-L109C14) which is passed to [`conditional_write()`](https://github.com/isehle/bye_splits/blob/482e84cd1bab79fa1fe7ce6f4a1f82aff869c3b2/bye_splits/production/submit_scripts/job_submit.py#L129). Before introducing the `raise ValueError`, `grab_most_recent()` would have returned `None`. In that case, the `v0` file would be written [here](https://github.com/isehle/bye_splits/blob/482e84cd1bab79fa1fe7ce6f4a1f82aff869c3b2/bye_splits/utils/job_helpers.py#L82). So introducing the `ValueError` in `grab_most_recent()` was actually a breaking change, when it was meant to be a bug fix. With `return None` in the `else` statement, `conditional_write()` should work as expected.

bye_splits/production/submit_scripts/job_submit.py
```diff
@@ -0,0 +1,221 @@
...
+        self.batches      = self.batch.setup_batches()
+
+    def _write_arg_keys(self, current_version):
+        """Writes the line containing the argument
+        names to the buffer list."""
+        
+        arg_keys = "Arguments ="
+        for arg in self.args.keys():
+            arg_keys += " $({})".format(arg)
+        arg_keys += "\n"
```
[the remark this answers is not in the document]
isehle: Well that's the point of the configuration file; it's up to the user to modify `config.yaml` with the arguments and values that `script` accepts.

bye_splits/production/submit_scripts/job_submit.py
```diff
@@ -0,0 +1,221 @@
...
+        """
+        if "particles" in self.args.keys(): self.args["particles"] = self.particle
+        arg_keys = ", ".join(self.args.keys())
+        arg_keys = "queue " + arg_keys + " from (\n"
+        current_version.append(arg_keys)
+        for batch in self.batches:
+            sub_args = list(self.args.keys())[1:]
+            arg_vals = [self.args[key] for key in sub_args]
+            all_vals = ["{}".format(batch).replace(", ", ";")]+arg_vals
+            all_vals = ", ".join(all_vals) + "\n"
```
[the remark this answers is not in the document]
isehle: Good catch, thank you! As for casting back to the original type, I think this should be handled directly by `argparse` when you set the `type` variable, see for example [here](https://github.com/bfonta/bye_splits/blob/156fa37eb4aebe00ce65cd6b9bb7517091c88431/bye_splits/production/produce.py#L181). Maybe @bfonta knows more, though.

README.md
isehle: Yes, that's already written and included in the latest commit (not pushed yet).

README.md
isehle: Yes yes I'll mention the skimming procedure application too!

As for running it on `EOS`, have you confirmed that you have access to `/eos/user/b/bfontana/FPGAs/new_algos/photons_0PU_bc_stc_hadd.root` just inside of your terminal? I do indeed remember using files on `EOS` being a major headache, however. I'll look into the problem further, but for now can you see if adding `/opt/exp_soft/cms/t3/eos-login -username $USER -init` to your `script` fixes the issue? Obviously this assumes that you have the same LLR/CERN username, which I don't, so if it's not replace `$USER` directly.

bye_splits/production/submit_scripts/job_submit.py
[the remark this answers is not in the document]
bfonta: I partially agree with both comments: I would add a check that guarantees the argument in the configuration file is supported by the script, raising an exception otherwise. This would make debugging much quicker.

In addition, I noticed the ```_write_arg_keys``` function is called only once, and it modifies the list being passed. When you define a function, you ideally want to make it crystal clear what it is supposed to do. In this case, the only way to know is to inspect it directly. This defeats the improved structuring that splitting the code into self-contained functions usually brings, since you force the reader to search for the function, decreasing the code's readability. In my opinion, given that you only use the function once, it would be better to include it directly where it is now being called.

I also suspect ```if len(self.args) > 0:``` is not needed, since ```Arguments =``` would be correct syntax even without arguments, but please correct me (I haven't tested it).

bye_splits/production/submit_scripts/job_submit.py
```diff
@@ -0,0 +1,221 @@
...
+            [0.01;0.02], photon
+        )
+        correctly assigns radius="[0.01, 0.02]", particle="photon"
+        """
+        if "particles" in self.args.keys(): self.args["particles"] = self.particle
+        arg_keys = ", ".join(self.args.keys())
+        arg_keys = "queue " + arg_keys + " from (\n"
+        current_version.append(arg_keys)
+        for batch in self.batches:
+            sub_args = list(self.args.keys())[1:]
```
bfonta, line 86: IIUC, ```[1:]``` is to differentiate the cluster radius from other parameters. This will not work in general, so such be adapted.

bye_splits/production/submit_scripts/job_submit.py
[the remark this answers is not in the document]
bfonta: I also think ```type``` handles it, but this would require testing.

bfonta (commented):
From @mchiusi comments, I realized there are many tiny details that might go wrong once someone else tries the job submission, so I should also test it on my side.
Would it be ok to include a quick recipe on the README in order to run a dummy test? You could add a dummy script that prints a parameter to a file (HTCondor ```*.out``` and ```*err``` files, for instance), and I would try to run it on my side by defining myself an array of parameters following your instructions. This would mimic running someone else's script with some parameters of my choice.

README.md
mchiusi: Yes from my terminal I have access to this file. I will try to run it adding the script you suggested directly in the script and I will let you know.

bye_splits/production/submit_scripts/job_submit.py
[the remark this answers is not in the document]
isehle: Good points, thank you. The functionality of `_write_arg_keys` has been moved out of a dedicated function and into `prepare_multi_job_condor` directly. I've written a quick argument checker which looks like:
```
    def _check_script_arguments(self):
        script_args = subprocess.run([self.script, "--args"], capture_output=True, text=True)
        if script_args.returncode != 0:
            result = script_args.stdout.strip().split("\n")
        else:
            result = []
        
        assert(result == list(self.args.keys())), "Provided arguments {} do not match accepted arguments {}.".format(list(self.args.keys()), result)
 ```
 This requires that any passed `script` has a named argument `--args` that would print out the names of the arguments:
 ```
         --args)
            echo "radius"
            echo "particles"
            echo "pileup"
            exit 1;;
```
If you have any ideas on a more flexible way of doing this let me know, or if you think the name should be changed.
I'll additionally write a dummy script that takes no arguments and determine if everything works as expected if we drop the `if len(self.args) > 0` check.

bye_splits/production/submit_scripts/job_submit.py
```diff
@@ -0,0 +1,221 @@
...
+            arg_vals = [self.args[key] for key in sub_args]
+            all_vals = ["{}".format(batch).replace(", ", ";")]+arg_vals
```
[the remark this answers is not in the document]
isehle: Because `batch` is a list, not a string. However, that makes me realize that `str(batch).replace(", ",";")` would be cleaner. Changed now.

bye_splits/production/submit_scripts/job_submit.py
isehle: Replaced with:
```
sub_args = list(self.args.keys())
sub_args.remove(self.iterOver)
```
For some reason, `sub_args = list(self.args.keys()).remove(self.iterOver)` returns `None`, so it must be done in two lines. This should be flexible enough to work generally, as the distinction is between constant parameters and iterated parameters.

bfonta: ```remove``` changes the list in-place

isehle:
The latest commit allows you to pass the Python files, i.e. `run_cluster.py` or `produce.py` for example, as `script` directly. The `executable` written by `write_exec_file()` in `job_submit.py` and set in the `.sub` file now calls the Python function directly on the correct batches of arguments. This has been tested and has worked well, but I still need to address a few of the other comments before this is ready to be merged.

bye_splits/production/submit_scripts/job_submit.py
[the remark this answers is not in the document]
isehle: Adding the assertion that the arguments in `config.yaml` match the arguments accepted by `<script>` is proving quite tricky. If either of you have some ideas on how this can be done generally and robustly please let me know, but if not this might deserve to have a dedicated PR.

bye_splits/production/submit_scripts/job_submit.py
isehle: It's currently only possible to iterate over a single parameter. If you would like I can work on extending it to iterating over multiple parameters, but it would complicate the batching considerably. The batches are currently subsets of the list of the iterated parameter, i.e. `[1,2,3,4,5] --> [[1,2], [3,4], [5]]` for example. This means that the `.sub` file would have a section like:
```
queue val from (
    [1,2],
    [3,4],
    [5]
)
```
Multiple iterated parameters would mean choosing which argument should be batched in this way, and which should create a new submission file/job, because as far as I understand, you can't create a `.sub` file with a `queue` statement within another `queue` statement.

bye_splits/production/submit_scripts/job_submit.py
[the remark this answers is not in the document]
isehle: Tested, `type` handles things as expected.

isehle:
Everything should be accounted for in the latest commit except for the things in the three remaining unresolved conversations. Personally I think these three things (argument verification, EOS files, and multiple iteration params) should wait for a dedicated PR because they're all fairly complicated. I'm happy to discuss more, but if you agree and everything else looks good, please resolve these remaining conversations.

bye_splits/production/submit_scripts/job_submit.py
bfonta: You could always rewrite things for the following (or similar) to work:

```shell
queue val1, val2 (
  1, 2
  3, 4
  ...
)
```

but I agree it increases the scope of the PR and can be addresses in a different PR in the case this is needed.

bye_splits/production/submit_scripts/job_submit.py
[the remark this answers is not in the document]
bfonta: When a script runs from the CLI you can access its required arguments via ```--help```, which means the information can be retrieved. I think this is important for this PR, but we can discuss.

bfonta:
@isehle Regarding argument verification, here goes my proposed solution.

You could run the script separately via the OS and parse its output. This is messy, it requires some non-python interface (even if within python), and it will be hard to generalize.
The alternative is to do everything within Python; the compromise is to slightly change the way we optional commands are now being defined. I propose we start encapsulating them into a function (a class would be more correct but an overkill for our current use-case).

You can find a running example [here](https://replit.com/join/qumsswonvk-balves16). If the solution suits you, I propose you add it to the dummy script, to your own script and to ```produce.py```, keeping all other scripts as they are. The reason is that these three scripts would be needed for testing. We can modify the others later on if really needed.

mchiusi:
I like Bruno's proposal to encapsulate the arguments of each script in a function so that they can be called from outside the script. In this way you could easily verify which are the needed arguments to launch a specific script.

I would also suggest adding the possibility to specify command-line arguments when running the job_submit.py script. The final result I have in mind is, in the case of one wants to run the `produce.py` script, something like this:
```
python bye_splits/production/submit_scripts/job_submit.py --script produce.py --particles photons
```
The command-line arguments should take precedence over those specified in `config.yaml`, which will only be read in case they are not specified via the command line.

Let me know what you think, and if you'd like, we can discuss it further!

isehle:
@bfonta, thanks for the suggestion and head start! I figured that it's probably best to start with a class even though I agree that it's overkill for now—it's just going to be easier to extend in the future, I imagine. You can find my approach [here](https://replit.com/join/dcthswumrm-ehle), where you can run `python run_script.py` which uses the `Arguments` class in `arg_script.py` to first verify that the arguments are expected by `test_script.py`, and then adds them and runs the script.
@mchiusi, I agree that it would be a nice idea to allow command line arguments at the level of `job_submit.py`, though I need to think a bit further on the best way to implement them—I have some ideas but it isn't trivial, so I'll keep you both updated.

isehle:
The class has been updated so that it now will take command line arguments with precedence over the default arguments, and everything is handled directly by the class itself so you can simply use `Arguments.run_script(<arg_dict>)` and (potential) replacement, verification, and running will be done behind the scenes. You can see the implementation at the same link as before.

isehle:
The latest commit should take care of all argument adding, combining (default+command line), and verification. Tested all combinations on the dummy script, and ran a few combinations with `run_cluster.py`.  As expected, it does indeed require changing the original script(s). Arguments must be defined with a dictionary titled `arg_dict` and added using `Arguments.add_args(arg_dict)`. The `keys` should be the argument name (including `--`, this is required for now), and the `values` should be their own dictionaries containing `**kwargs` that are accepted by `arparse.add_arguments()`.
Example:
```
arg_dict = {
    "--float_arg": {
        "help": "Dummy float",
        "type": float,
        "required": True
    },
    "--str_arg": {
        "help": "Dummy string.",
        "type": str,
        "required": True,
    },
    # Generic argument will be interpreted as a string
    "--gen_arg": {
        "help": "Dummy generic.",
        "required": False,
        "default": None
    },
    "--flag": {
        "help": "Dummy flag number 2.",
        "action": "store_true",
        "required": False
    }
}

def dummy(pars):
    """Dummy script for testing HT Condor job submission.
    Prints passed argument key/vals and the type of <val>."""

    for key, val in pars.items():
        print("Passed {} for --{}, read as type {}.".format(val, key, type(val)))

if __name__ == "__main__":
    args = Arguments(script=__file__)
    FLAGS = args.add_args(description="A dummy script to submit.", arg_dict=arg_dict)
    dummy(FLAGS)
```

isehle:
I suggest playing with `tests/submission/dummy_run.py` to get a feel for how Arguments handles things. A few things to note:
1) It is _not_ necessary for the dictionary in the running script to have any specific name. It must be `arg_dict` in the script that accepts the arguments so that these can be read by the `Arguments` class and verified against what's passed in the running script. The name in the running script however has no effect.
2) Try removing a required argument in the run script argument dictionary.
3) Try changing arguments from the command line.
4) Try adding an argument that isn't accepted by the "fundamental" script.

